PR #28, opened 2015-02-04
Prevent duplicate form submissions with JavaScript

rjkip:


rjkip:
@DRvanR this

DRvanR:
As the same must be used on all forms (so also RA/GW), might be an idea to just include it in https://github.com/SURFnet/Stepup-bundle and include that JS in the main templates in the apps?

rjkip:
weeee @DRvanR

DRvanR:
bleergh

DRvanR:
oh, eeeh, and saddas

PR #32, opened 2015-02-09
Reset the SAML flow when an exception occurs

rjkip:


rjkip:
knot
@DRvanR

src/Surfnet/StepupSelfService/SelfServiceBundle/Security/Firewall/SamlListener.php
```diff
@@ -74,6 +74,17 @@ public function __construct(
 
     public function handle(GetResponseEvent $event)
     {
+        try {
+            $this->tryHandle($event);
```
DRvanR: `handleEvent`? tryHandle sounds so meh...

DRvanR:
Minor nitpick-of-mehness, otherwise :+1:

DRvanR:
(same can be applied in RA btw, GW not so sure)

PR #35, opened 2015-02-11
Remove the mobile phone number from URL

rjkip:
Terminology explanation for [this diff line](https://github.com/SURFnet/Stepup-SelfService/commit/93b7322b1b6ab71500255f24d83f98f7b0b84c0c#diff-ae1f51068dfaf8da2e870c6fb62e5249L36): 'taking' implies a side effect (reading and removing the phone number) as opposed to 'getting', which implies no side effects.

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SmsSecondFactor/SessionChallengeStore.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SmsSecondFactor/SessionChallengeStore.php
@@ -42,7 +42,7 @@ class SessionChallengeStore implements ChallengeStore
         $this->sessionKey = $sessionKey;
     }
 
-    public function generateChallenge()
+    public function generateChallenge($phoneNumber)
     {
         $randomCharacters = function () {
             $chr = rand(50, 81);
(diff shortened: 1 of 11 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SmsSecondFactor/SessionChallengeStore.php
```diff
@@ -52,18 +52,16 @@ public function generateChallenge()
...
 
         return $challenge;
     }
 
-    public function verifyChallenge($challenge)
+    public function takePhoneNumberMatchingChallenge($challenge)
     {
         $challenge = strtoupper($challenge);
-        $expectedChallenge = $this->session->get($this->sessionKey);
+        $phoneNumber = $this->session->remove($this->sessionKey . '/' . $challenge);
```
DRvanR: Hrms, seems the SessionBagInterface doesn't contain remove :panda_face: 

Nitpick: Why `takeXXX` by the way? Doesn't seem to match with the `generateChallenge()` naming style...

DRvanR:
Hmms, seems I rambled on a bit... sorry

Could you add some tests covering the `SessionChallengeStore`? That seems to do the primary checking now depending on whether or not some key exists in the session, hence the request. 
Might also be good to test that in the `SmsSecondFactorService` the `CommandService::execute()` is only called with a ` new ProvePhonePossessionCommand()` if the `SessionChallengeStore` returns a phonenumber. Rather not do those kind of checks, but seeing that the validation is now done implicit (whether or not a particular sessionkey exists) rather than explicit (does challenge match entered challenge) by some service this might be warranted.

Wondering about possible security implications for brute-forcing this: one could go on with attempting with different OTPs until the one in the session is found. Might be best to limit to three attempts or something and always remove after that. Dunno

All in all I've got the feeling the implementation could be a bit more explicit and robust than this. Thoughts?

DRvanR:
W.r.t. `takeXXX`:

ugh I don't read descriptions :wink: - still stands that both setting and removing are implemented as side-effects of something else, but are actually rather critical to the process as a whole....

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SmsSecondFactor/SessionChallengeStore.php
rjkip: I'm type-hinting against a `SessionInterface`, which contains remove.

rjkip:
> still stands that both setting and removing are implemented as side-effects of something else, but are actually rather critical to the process as a whole....

I tried to create some sort of coat check where you say "here's my phone number, give me a secret," and when you come back later with your secret, you get your phone number back ;)

PR #37, opened 2015-02-17
Hide the phone number in the URL. Allow three codes. Add expiry time.

rjkip:
* [Do not show 06-nr in plain text in address bar browser](https://www.pivotaltracker.com/story/show/86103512)
 * [As a Registrant, SMS challenges I receive expire after X minutes.](https://www.pivotaltracker.com/story/show/83691530)
 * [As a Registrant, I may only request three SMS challenge retries](https://www.pivotaltracker.com/story/show/83691800)


rjkip:
:memo: @DRvanR

build-pre-commit.xml
```diff
@@ -137,7 +148,7 @@
     </target>
 
     <target name="php-license" depends="get-changeset.php-spacesep" if="changeset.php.notempty">
-        <exec executable="/bin/grep" failonerror="true" outputproperty="php-license.grep.output">
+        <exec executable="/bin/grep" failonerror="false" outputproperty="php-license.grep.output">
```
DRvanR: Why no fail anymore? :cry:

rjkip: It still fail, sir. When it is set to true, the build just fails on that step. When false, it shows a list of files without licence headers.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/Registration/SmsController.php
```diff
@@ -45,9 +46,17 @@ public function sendChallengeAction(Request $request)
             /** @var SmsSecondFactorService $service */
             $service = $this->get('surfnet_stepup_self_service_self_service.service.sms_second_factor');
 
-            if ($service->sendChallenge($command)) {
+            try {
+                $smsSendingSucceeded = $service->sendChallenge($command);
+            } catch (TooManyChallengesRequestedException $e) {
+                $form->addError(new FormError('ss.prove_phone_possession.challenge_request_limit_reached'));
+
+                return ['form' => $form->createView()];
```
DRvanR: Maybe also show how many SMS's they still can request? Might be worth a discussion at the least...

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SmsSecondFactor/SessionChallengeHandler.php
```diff
@@ -0,0 +1,104 @@
...
+
+    public function requestOtp($phoneNumber)
+    {
+        $randomCharacters = function () {
+            $chr = rand(50, 81);
+
+            // 9 is the gap between "7" (55) and "A" (65).
+            return $chr >= 56 ? $chr + 9 : $chr;
+        };
+        $otp = join('', array_map('chr', array_map($randomCharacters, range(1, 8))));
```
DRvanR: If you move the `chr` call into the `randomCharacters` function (and call it `randomCharacter` :wink: ) you can skip an `array_map` call. Seems that might make it a bit more readable... Thoughts?

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/Registration/SmsController.php
rjkip: Let's discuss in the stand-up.

rjkip:
:memo: @DRvanR

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SmsSecondFactor/Otp.php
```diff
@@ -0,0 +1,111 @@
...
+        return join('', array_map($randomCharacters, range(1, 8)));
+    }
+
+    /**
+     * @param string $otpString
+     * @param string $phoneNumber
+     * @param DateInterval $expiryInterval
+     * @return Otp
+     */
+    public static function create($otpString, $phoneNumber, DateInterval $expiryInterval)
```
DRvanR: Would fold the now given `$otpString` into this method with an accessor to retrieve the Otp (or a `__toString`), or move the `::generate()` to a different class. The call is now 

```
$otp = Otp::generate()
$otp = Otp::create($otp, ...)
```

seems a bit redundant...

Thoughts?

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SmsSecondFactor/SessionSmsVerificationStateHandler.php
```diff
@@ -0,0 +1,107 @@
...
+        return $this->otpRequestMaximum;
+    }
+
+    public function verify($otp)
+    {
+        /** @var SmsVerificationState|null $state */
+        $state = $this->session->get($this->sessionKey);
+
+        if (!$state) {
+            return OtpVerification::matchExpired();
```
DRvanR: If there is no state it expired? Not sure about this...

DRvanR: Think noMatch might be better?

DRvanR:
Two small remarks, looks good otherwise!

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SmsSecondFactor/SessionSmsVerificationStateHandler.php
rjkip: It's just that if somehow a registrant returns to this page without a session *and* has received a code (eg. browser crashes after receiving a code), a *no-match* will confuse the registrant, whereas an expiration message will prompt him/her to request a new code. If this is not acceptable we should imo prompt the user to send a new code.

DRvanR: What if the user returns to this page without a session and without having received a status-code (for instance by using, however stupid, a bookmark)? There is no up front check whether or not the user has visited the previous page and thus may or may not have received a status code.

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SmsSecondFactor/Otp.php
rjkip: I disliked the idea of always being able to extract the secret from the OTP (`$otp->verify($otp->getOtp())`). Considering it doesn't require "extensive" refactoring, I'd agree to moving generate to a different class.

DRvanR: :+1:

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SmsSecondFactor/SessionSmsVerificationStateHandler.php
rjkip: Reading all this it seems best to redirect to the send-code page if there is no `SmsVerificationState` and inform the registrant he/she has to request a new code.

DRvanR: agreed, sensible _and_ practical :wink:

DRvanR:
:+1: Merge away!

rjkip:
But, but, my GIF...

bleep-blorp

PR #41, opened 2015-02-23
Render informative error screens in case of the below resp. statuses.

rjkip:
- AuthnFailed - IdP failed to authenticate user
 - * - All other cases where preconditions weren't met

The reason the response is rendered and set from the SamlListener and
not from a kernel listener is the fact that the firewall errors out
when the listener doesn't set a token.


rjkip:
pokemans

DRvanR:
wink :+1:

PR #45, opened 2015-03-09
Replace own commands with MiddlewareClientBundle Commands

DRvanR:
Depends on https://github.com/SURFnet/Stepup-Middleware-clientbundle/pull/18

- [ ] Update composer.json to point to develop after merging aforementioned PR


DRvanR:
@rjkip

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/YubikeySecondFactor/ProofOfPossessionResult.php
```diff
@@ -91,7 +132,7 @@ public function didOtpVerificationFail()
     /**
      * @return boolean
      */
-    public function didProofOfPossessionFail()
+    public function didProofOfPossessionCommandFail()
```
rjkip: Is it relevant to the consumer of the service that it was in fact the execution of a command that failed? Imo the consumer is merely asking the possession to be proven using whatever method the service sees fit.

rjkip:
Nice `ProofOfPossessionResult` named constructors. :+1:  Do what you want with my question.

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/YubikeySecondFactor/ProofOfPossessionResult.php
DRvanR: It was/is the structure and apparent intent of the Result objects to communicate success/failure and if failed, why it failed. This in order to be able to provide correct feedback to the user (at least, that is what I gathered from it).
This name was changed in order to better reflect the fact that the Proof of Possession as given by the User was in fact successful, however there was a system error (see the named constructor and inner logic of the class). 
I'm not to keen on how these classes are implemented in general though...
Will add a chore to revisit them later as we have more pressing matters to tend to :smiley:

PR #66, opened 2015-04-16
Display sign-out button when user is authenticated

rjkip:
d


rjkip:
Changes now ensure the token is persisted between requests by Symfony's default behaviour (using the ContextListener).

Nelmio SecurityBundle seems to be unable to sign and encrypt the session cookie. I am unable to ascertain where/when the session cookie is set; it is not in the Response object on `Response#send()`. This is why I have disabled the signing and encryption of all cookies.

PR #71, opened 2015-06-02
Include preferred locale when creating an identity

rjkip:


rjkip:
:hankey: @DRvanR

DRvanR:
:+1:

PR #74, opened 2015-06-11
Move form errors from 'validators' domain to 'messages' and translate other missing translation keys

rjkip:


rjkip:
gi

@DRvanR

DRvanR:
:+1:

PR #92, opened 2015-09-29
After revoking a U2F second factor, revoke U2F device registration from U2F verification server

rjkip:


rjkip:
@DRvanR 

df9

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/U2fService.php
```diff
@@ -205,6 +208,69 @@ public function verifyRegistration(VerifyU2fRegistrationCommand $command)
...
+        if ($hasErrors && $statusCode >= 400 && $statusCode < 600) {
+            $this->logger->critical(
+                sprintf(
+                    'U2F registration revocation failed; HTTP %d with errors "%s"',
+                    $statusCode,
+                    join(', ', $result['errors'])
+                )
+            );
+
+            return RegistrationRevocationResult::apiError();
```
DRvanR: This is unrecoverable, the Middleware has already acted: the actual revocation has been executed.
Failure of the GW api is fault-tolerant: it will not cause issues in the future as a new registration will issue a different keyhandle (at least, to my understanding :wink:).

Hence I think all GW revoke API errors should be properly logged, but not cause errors to be presented to the user.

rjkip: This should not cause any errors to be presented to the user. Actually, I don't even use the revocation result. I still wanted to communicate the result, though.

DRvanR: uuuuugh, somehow my mind converted all `return` statements to `throw` statements :( mi scusi

rjkip: Can I freely interpret this as a thumb directed away from the earth's centre?

DRvanR: why yes, yes you can :+1:
--
ibuildings
WEB & MOBILE APP DEVELOPMENT

Daan van Renterghem
Software Engineer

phone (direct): +31 30 75 51 590
skype: dvrenterghem-ibuildings
e-mail: dvrenterghem@ibuildings.nl
phone: +31 88 00 24 000
www.ibuildings.nl

To all agreements with Ibuildings.nl BV, our general terms and conditions
<http://www.ibuildings.nl/algemene_voorwaarden> apply.

PR #93, opened 2015-09-30
Prevent attackers from submitting illegal return URLs

rjkip:
Before this change, a return URL of https://selfservice.tld@attack.tld
was acceptable, because it started with our app's HTTP scheme and
host.

Addresses [Security Audit 5.2.12](https://www.pivotaltracker.com/story/show/104421194).


DRvanR:
:+1:

PR #98, opened 2016-06-06
Make Tiqr app descriptions into links

relaxnow:
https://www.pivotaltracker.com/projects/1163646/stories/113748599

@rjkip can you review?

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/views/Registration/partial/secondFactor.html.twig
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/views/Registration/partial/secondFactor.html.twig
@@ -14,7 +14,7 @@
         <hr>
         <p>
             {{
-                ('ss.registration.selector.'~type~'.description')|trans(
+                ('ss.registration.selector.'~type~'.description')|trans|e|replace(
                     {
                         '%tiqr_android_link_start%': '<a href="'~tiqrAppAndroidUrl~'">',
                         '%tiqr_android_link_end%': '</a>',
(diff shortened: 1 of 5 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/views/Registration/partial/secondFactor.html.twig
```diff
@@ -13,7 +13,16 @@
         </p>
         <hr>
         <p>
-            {{ ('ss.registration.selector.'~type~'.description')|trans }}
+            {{
+                ('ss.registration.selector.'~type~'.description')|trans(
+                    {
+                        '%tiqr_android_link_start%': '<a href="'~tiqrAppAndroidUrl~'">',
```
rjkip: Hmm, good to know these parameters aren't escaped :sweat_smile:.

:ship:

relaxnow: They are escaped, see ```|raw``` on: https://github.com/SURFnet/Stepup-SelfService/pull/98/files#diff-43e7ab398b2e3d7574c3dd5c0f568be9R24

There appears to be no way around ```raw``` 😢

rjkip: Good, I thought I was going mad. Well in that case, if you want to do the "correct" thing, there's also [`|replace`](http://twig.sensiolabs.org/doc/filters/replace.html) :stuck_out_tongue: (in other words: `|trans|e|replace(...)|raw`)

rjkip: But feel free to :ship:

relaxnow: What is the advantage of ```replace``` over ```trans``` arguments?

relaxnow: Ah manual escaping with ```e```

PR #103, opened 2016-06-13
Line out form element and button for SMS send-challenge

relaxnow:
screen shot 2016-06-13 at 11 25 20


rjkip:
That looks kinda weird. I think the select/input locations and sizes were fine, just that the button was too far to the right and should be moved left. What about making `subscriber` fill its container and then making the form narrower (with columns offsets and smaller columns widths maybe)?

rjkip:
Sorry, that's exactly what they requested. :+1:

arothuis:
It looks like requested. :+1: 

It's definitely out of scope for now, but I would like to note that the concepts could be grouped via [inline-forms][inline-form] or [input-groups][input-groups] as the form elements here are related.

[inline-form]: http://getbootstrap.com/css/#forms-inline
[input-groups]: http://getbootstrap.com/components/#input-groups

relaxnow:
@arothuis nice, I'm unfamiliar with that bootstrap feature I'll have to check it out next time.

PR #109, opened 2016-08-05
Show RAA contact information if it has been configured, otherwise not

arothuis:
This should also be tested more extensively and can be done in conjunction with Middleware's changes regarding this option.


DRvanR:
Seems that the logic could be simplified, but other than that, :+1:

DRvanR:
:+1:

PR #110, opened 2016-08-10
Refined firewall with explicit session timeout and prefetch handling

DRvanR:
Includes both https://www.pivotaltracker.com/story/show/115162333 and https://www.pivotaltracker.com/story/show/121296389 for the SelfService.


DRvanR:
I have left the functional test out of this PR, this is on purpose. They did not seem to be willing to run in any other env than dev, which kinda defeats the purpose. Will try to fix those ASAP. In the mean time, it has been fully manually tested.

Different scenario's:

#### Expired through absolute lifetime

- log in, let maximum absolute lifetime for session expire (for sanity, configure a short lifetime in parameter.yml)
- attempt to visit page
- click login button, or visit known URL (e.g. `/`), and be prompted to log in again (through EB).

#### Expired through relative lifetime

- log in, let maximum relative lifetime for session expire (for sanity, configure a short lifetime in parameter.yml)
- attempt to visit page
- click login button, or visit known URL (e.g. `/`), and be prompted to log in again (through EB).

#### Allow prefetched request by introducing ability to send secondary AuthnRequest

- visit a page, be prompted to log in through EB (simulates prefetch by browser)
- do not log in, but visit the page again, be prompted to log in through EB with a new AuthnRequest.
- log in
- return successfully to app, logged in and all.

Of course all standard behavior of a firewall is still intact.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SecurityController.php
```diff
@@ -0,0 +1,48 @@
...
+{
+    public function sessionExpiredAction(Request $request)
+    {
+        $redirectToUrl = $this
+            ->get('self_service.security.authentication.session.session_storage')
+            ->getCurrentRequestUri();
+
+//        // Return URLs generated by us always include a path (ie. at least a forward slash)
+//        // @see https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Request.php#L878
+//        $domain = $request->getSchemeAndHttpHost() . '/';
```
arothuis: What is the purpose of this commented out code?

app/config/parameters.yml.dist
```diff
@@ -54,3 +54,6 @@ parameters:
 
     tiqr_app_android_url: https://play.google.com/store/apps/details?id=org.tiqr.authenticator&hl=en
     tiqr_app_ios_url: https://itunes.apple.com/us/app/tiqr/id430838214?mt=8&ls=1
+
+    session_max_absolute_lifetime: 3600 # 1 hours * 60 minutes * 60 seconds
```
arothuis: I think the parameter names are clear enough for us, but maybe these parameters (and the configuration in general) could be documented a bit more (some comments in the dist would suffice, I think). I will make a tech debt issue for this.

arothuis:
Looks good, also nice clean-up. Thanks for the thorough explanation in your comment regarding testing.

Minor comments, otherwise 👍

pmeulen:
Should fix the prefetching issue described in https://www.pivotaltracker.com/story/show/122115911 as well.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SecurityController.php
DRvanR: Heh, should be removed. At first the return URL was a GET parameter, which have to be checked so that people are not redirected to evil-domain.org. Now it is a session parameter, and therefor fully under our control. Can be removed.

src/Surfnet/StepupSelfService/SelfServiceBundle/Security/Authentication/Handler/ExplicitSessionTimeoutHandler.php
```diff
@@ -0,0 +1,144 @@
...
+        $this->logger                      = $logger;
+    }
+
+    public function process(GetResponseEvent $event)
+    {
+        if ($this->tokenStorage->getToken() !== null
+            && !$this->sessionLifetimeGuard->sessionLifetimeWithinLimits($this->authenticatedSession)
+        ) {
+            $invalidatedBy = [];
+            if ($this->sessionLifetimeGuard->sessionLifetimeWithinAbsoluteLimit($this->authenticatedSession)) {
```
arothuis: The session will be logged as invalidated by the absolute limit when the lifetime is within the limit instead of outside the limit, I think this check (and the one below) should be inverted.

src/Surfnet/StepupSelfService/SelfServiceBundle/Security/Authentication/Session/SessionStorage.php
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * @param SessionInterface $session
+     */
+    public function __construct(SessionInterface $session)
+    {
+        $this->session = $session;
+    }
+
+    public function logAuthenticationMoment()
```
arothuis: Other than in tests, I don't see this being called anywhere.

I think this should be done upon (first) authentication in order to check whether the session lifetime exceeds the absolute limit or not.

src/Surfnet/StepupSelfService/SelfServiceBundle/Security/Authentication/Handler/ProcessSamlAuthenticationHandler.php
```diff
@@ -159,6 +159,7 @@ public function process(GetResponseEvent $event)
                 return;
             }
 
+            $this->authenticatedSession->logAuthenticationMoment();
```
DRvanR: good catch, checked it and indeed had a commit still local that added this (rebased a bit too much before pushing :flushed: ) - hidden between adding lots of debug log messages haha.

PR #111, opened 2016-10-11
Implement locale cookie listener from StepupBundle into SelfService

arothuis:
[113611545](https://www.pivotaltracker.com/story/show/113611545)

This implements the locale cookie listener from the StepupBundle into SelfService.

For its RA counter-part, see: [Stepup-RA #123](https://github.com/SURFnet/Stepup-RA/pull/123).


arothuis:
@rjkip

PR #113, opened 2016-11-02
SelfService: Multi locale logout redirect url

arothuis:
[131975453](https://www.pivotaltracker.com/story/show/131975453)

This PR adds a way to configure a logout redirect URL.

#### Related
SURFnet/Stepup-RA#129
SURFnet/Stepup-RA#128
SURFnet/Stepup-SelfService#112


arothuis:
This feature has been deemed unnecessary at this time as a single URL suffices in practice.

PR #117, opened 2017-02-13
Apply Allowed Second Factors in SS

arothuis:
[119815553](https://www.pivotaltracker.com/story/show/119815553)

OpenConext/Stepup-Middleware-clientbundle#49 and OpenConext/Stepup-Middleware#179 should be reviewed and merged first.

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RegistrationController.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RegistrationController.php
@@ -34,14 +34,17 @@ class RegistrationController extends Controller
         $institutionConfigurationOptions = $this->get('self_service.service.institution_configuration_options')
             ->getInstitutionConfigurationOptionsFor($this->getIdentity()->institution);
 
-        $allowedSecondFactors = array_intersect(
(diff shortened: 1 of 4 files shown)


arothuis:
Before merging: update the Middleware client bundle in composer.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RegistrationController.php
```diff
@@ -31,11 +31,17 @@ class RegistrationController extends Controller
      */
     public function displaySecondFactorTypesAction()
     {
-        $enabledSecondFactors = $this->getParameter('ss.enabled_second_factors');
+        $institutionConfigurationOptions = $this->get('self_service.service.institution_configuration_options')
+            ->getInstitutionConfigurationOptionsFor($this->getIdentity()->institution);
+
+        $allowedSecondFactors = array_intersect(
```
DRvanR: now I get why it is an String[] rather than SecondFactorType[] 😉 

Nice and proper use of array_intersect! Wish there was a more elegant solution but for now this'll suffice.

PR #118, opened 2017-02-23
Add second factor test button

nicwortel:
Part of https://www.pivotaltracker.com/story/show/79436876, will be followed by PR's to implement the test controller action and to handle the authentication response.

PR #120, opened 2017-02-27
Handle second factor test response

nicwortel:
Part of https://www.pivotaltracker.com/story/show/79436876

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
@@ -23,6 +23,8 @@ use Surfnet\SamlBundle\Http\XMLResponse;
 use Surfnet\SamlBundle\SAML2\Response\Assertion\InResponseTo;
 use Surfnet\StepupBundle\Value\SecondFactorType;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Security\Core\Exception\AuthenticationException;
 
 class SamlController extends Controller
@@ -51,7 +53,7 @@ class SamlController extends Controller
                 )
             );
 
-            throw $this->createNotFoundException();
+            throw new NotFoundHttpException();
         }
 
         $loaResolutionService         = $this->get('surfnet_stepup.service.loa_resolution');
@@ -86,7 +88,7 @@ class SamlController extends Controller
                 'Received an authentication response for testing a second factor, but no second factor test response was expected'
             );
 
-            throw $this->createAccessDeniedException('Did not expect an authentication response');
+            throw new AccessDeniedHttpException('Did not expect an authentication response');
         }
 
         $initiatedRequestId = $session->get('second_factor_test_request_id');
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
@@ -64,7 +64,7 @@ class SamlController extends Controller
             $loaResolutionService->getLoaByLevel($secondFactorType->getLevel())
         );
 
-        $this->get('session')->set('second_factor_test_mode', true);
+        $this->get('session')->set('second_factor_test_request_id', $authenticationRequest->getRequestId());
 
         $this->get('logger')->notice(
             sprintf(
@@ -82,7 +82,7 @@ class SamlController extends Controller
 
         $session = $this->get('session');
 
-        if (!$session->has('second_factor_test_mode')) {
+        if (!$session->has('second_factor_test_request_id')) {
             $this->get('logger')->error(
                 'Received an authentication response for testing a second factor, but no second factor test response was expected'
             );
@@ -90,7 +90,7 @@ class SamlController extends Controller
             throw $this->createAccessDeniedException('Did not expect an authentication response');
         }
 
-        $session->remove('second_factor_test_mode');
+        $session->remove('second_factor_test_request_id');
 
         $postBinding = $this->get('surfnet_saml.http.post_binding');
 
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
@@ -67,13 +67,7 @@ class SamlController extends Controller
         $this->get('session')->set('second_factor_test_request_id', $authenticationRequest->getRequestId());
 
         $logger = $this->get('surfnet_saml.logger')->forAuthentication($authenticationRequest->getRequestId());
-
-        $logger->notice(
-            sprintf(
-                'Sending authentication request with ID "%s" to the second factor test IDP',
-                $authenticationRequest->getRequestId()
-            )
-        );
+        $logger->notice('Sending authentication request to the second factor test IDP');
 
         return $this->get('surfnet_saml.http.redirect_binding')->createRedirectResponseFor($authenticationRequest);
     }
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
@@ -96,8 +96,6 @@ class SamlController extends Controller
 
         $postBinding = $this->get('surfnet_saml.http.post_binding');
 
-        $translator = $this->get('translator');
-
         try {
             $postBinding->processResponse(
                 $httpRequest,
@@ -105,12 +103,9 @@ class SamlController extends Controller
                 $this->get('surfnet_saml.hosted.service_provider')
             );
 
-            $session->getFlashBag()->add(
-                'success',
-                $translator->trans('ss.test_second_factor.verification_successful')
-            );
+            $session->getFlashBag()->add('success', 'ss.test_second_factor.verification_successful');
         } catch (Exception $exception) {
-            $session->getFlashBag()->add('error', $translator->trans('ss.test_second_factor.verification_failed'));
+            $session->getFlashBag()->add('error', 'ss.test_second_factor.verification_failed');
         }
 
         return $this->redirectToRoute('ss_second_factor_list');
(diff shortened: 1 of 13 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
```diff
@@ -18,29 +18,100 @@
...
+
+        $secondFactor     = $secondFactorService->findOneVetted($secondFactorId);
+        $secondFactorType = new SecondFactorType($secondFactor->type);
+
+        $authenticationRequest = $authenticationRequestFactory->createSecondFactorTestRequest(
+            $identity->nameId,
+            $loaResolutionService->getLoaByLevel($secondFactorType->getLevel())
+        );
+
+        $this->get('session')->set('second_factor_test_mode', true);
```
DRvanR: A bit more should be stored (and tracked) in the session; e.g. the request ID of the AuthnRequest.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
```diff
@@ -18,29 +18,100 @@
...
-        /** @var \SAML2_Assertion $assertion */
-        $assertion = $postBinding->processResponse(
-            $httpRequest,
-            $this->get('surfnet_saml.remote.idp'),
-            $this->get('surfnet_saml.hosted.service_provider')
-        );
+        $translator = $this->get('translator');
+
+        try {
+            $postBinding->processResponse(
```
DRvanR: it must be verified that the received response is in response to the sent request; for reference see [here][1]

[1]: https://github.com/OpenConext/Stepup-SelfService/blob/343d5ec8c9690e25a147bf0d054ff88fab4654b0/src/Surfnet/StepupSelfService/SelfServiceBundle/Security/Authentication/Handler/ProcessSamlAuthenticationHandler.php#L107

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
```diff
@@ -18,29 +18,100 @@
...
+        $secondFactorType = new SecondFactorType($secondFactor->type);
+
+        $authenticationRequest = $authenticationRequestFactory->createSecondFactorTestRequest(
+            $identity->nameId,
+            $loaResolutionService->getLoaByLevel($secondFactorType->getLevel())
+        );
+
+        $this->get('session')->set('second_factor_test_mode', true);
+
+        $this->get('logger')->notice(
```
DRvanR: The moment we have an AuthnRequest and thus a RequestID, we should use the [SamlAuthenticationLogger][1] (same when processing the response). For an example as how to use that, see [here][2]

[1]: https://github.com/OpenConext/Stepup-saml-bundle/blob/develop/src/Monolog/SamlAuthenticationLogger.php
[2]: https://github.com/OpenConext/Stepup-SelfService/blob/343d5ec8c9690e25a147bf0d054ff88fab4654b0/src/Surfnet/StepupSelfService/SelfServiceBundle/Security/Authentication/Handler/ProcessSamlAuthenticationHandler.php#L108

nicwortel: Is it correct that this logger can only be used when the request ID is set? In other words, the `SamlAuthenticationLogger` is invalid without a request ID and the default `logger` service should be used before we know the request ID?

nicwortel: Also, do you prefer to keep the request ID in the log message (as an argument for the `sprintf` call), or is it enough to add it to the logging context?

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
nicwortel: Should the second factor test flow also use an implementation of the `SamlAuthenticationStateHandler`? Or is it more appropriate to use the `session` directly?

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
```diff
@@ -18,29 +18,100 @@
...
+        try {
+            $postBinding->processResponse(
+                $httpRequest,
+                $this->get('self_service.second_factor_test_idp'),
+                $this->get('surfnet_saml.hosted.service_provider')
+            );
+
+            $session->getFlashBag()->add(
+                'success',
+                $translator->trans('ss.test_second_factor.verification_successful')
```
nicwortel: I noticed from the logs that this string is being translated twice: once here in the controller, and once more in the `session_flash` macro that is provided by the MopaBootstrapBundle. This triggers a warning, because the already translated string cannot be translated again.

However, when I remove the translation here, the string is no longer recognized as a translation key by the JMSTranslationBundle, so it will be removed from the result of the `extract-translations` script.

I have found a `translations.twig` file in the SelfServiceBundle which (I presume) exists to work around this problem. Is that correct? So then I could remove the translator here and add the translation keys to that file?

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
arothuis: The SamlAuthenticationLogger can only be used when the requestId is set.
It is enough to have it in the logging context.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
arothuis: Ideally we would want to keep the test implementation as close to the actual implementation. However, the intent and flow is different. Therefore, one would ideally create a wrapper for this specific case. For now, I'd use the session and register the improvement as technical debt.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
arothuis: Nice find. I think that should work. It is consistent with how other flash messages are done, see [SmsController.php (82)](https://github.com/OpenConext/Stepup-SelfService/blob/343d5ec8c9690e25a147bf0d054ff88fab4654b0/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/Registration/SmsController.php#L82) and [translations.twig (31)](https://github.com/OpenConext/Stepup-SelfService/blob/343d5ec8c9690e25a147bf0d054ff88fab4654b0/src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/views/translations.twig#L31)

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
nicwortel: Tech debt ticket created: https://www.pivotaltracker.com/story/show/141126081

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
```diff
@@ -18,29 +18,110 @@
...
-     * @Template
+     * @param string $secondFactorId
+     *
+     * @return \Symfony\Component\HttpFoundation\RedirectResponse
+     * @throws \Symfony\Component\HttpKernel\Exception\NotFoundHttpException
+     * @throws \Symfony\Component\Security\Core\Exception\AccessDeniedException
      */
+    public function testSecondFactorAction($secondFactorId)
+    {
+        $this->get('logger')->notice(
```
arothuis: You could extract `$this-get('logger')` to a variable as it is reused later on.
Also, this could fit on a single line.

The latter `$logger` could be renamed to `$samlLogger` or something else.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
```diff
@@ -18,29 +18,110 @@
...
+        if (!$secondFactorService->identityHasSecondFactorOfStateWithId($identity->id, 'vetted', $secondFactorId)) {
+            $this->get('logger')->error(
+                sprintf(
+                    'Identity "%s" tried to test second factor "%s", but does not own that second factor or it is not vetted',
+                    $identity->id,
+                    $secondFactorId
+                )
+            );
+
+            throw $this->createNotFoundException();
```
arothuis: Why not throw the exception itself?

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
```diff
@@ -18,29 +18,110 @@
...
+        $logger->notice('Received an authentication response for testing a second factor');
+
+        $session = $this->get('session');
+
+        if (!$session->has('second_factor_test_request_id')) {
+            $logger->error(
+                'Received an authentication response for testing a second factor, but no second factor test response was expected'
+            );
+
+            throw $this->createAccessDeniedException('Did not expect an authentication response');
```
arothuis: Same here; why not throw the exception itself?

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
```diff
@@ -18,29 +18,110 @@
...
+        }
+
+        $initiatedRequestId = $session->get('second_factor_test_request_id');
+
+        $logger = $this->get('surfnet_saml.logger')->forAuthentication($initiatedRequestId);
```
arothuis: You are reassigning the `$logger` here, you would probably want to rename this to prevent confusion.

nicwortel:
Note that I could not reproduce a situation where testing the second factor fails because the request ID differs from the one that is stored in the session (for instance by firing two requests in parallel), because the gateway also fails in such a scenario.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
```diff
@@ -18,29 +18,110 @@
...
+                        'Expected a response to the request with ID "%s", but the SAMLResponse was a response to a different request',
+                        $initiatedRequestId
+                    )
+                );
+
+                throw new AuthenticationException('Unexpected InResponseTo in SAMLResponse');
+            }
+
+            $session->getFlashBag()->add('success', 'ss.test_second_factor.verification_successful');
+        } catch (Exception $exception) {
```
arothuis: This would also catch an Exception if anything goes wrong in Gateway or client bundles (which is possible when sessions are somehow tainted during testing).

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
nicwortel: This is just a Symfony convenience method that creates a NotFoundHttpException with a default message, I guess it is a matter of personal preference.

nicwortel:
Includes changes from #118 and #119.

PR #122, opened 2017-03-27
Correctly implement Guzzle 6 interfaces

arothuis:
[126126187](https://www.pivotaltracker.com/story/show/126126187)

This correctly implements Guzzle 6 interfaces and uses the shared JsonHelper from the StepupBundle.


src/Surfnet/StepupSelfService/SelfServiceBundle/Service/U2fService.php
```diff
@@ -165,11 +167,11 @@ public function verifyRegistration(VerifyU2fRegistrationCommand $command)
             ],
         ];
 
-        $response = $this->guzzleClient->post('api/u2f/register', ['json' => $body, 'exceptions' => false]);
+        $response = $this->guzzleClient->post('api/u2f/register', ['json' => $body, 'http_errors' => false]);
```
nicwortel: Please do this for `createRegisterRequest()` as well

PR #125, opened 2017-06-07
Feature/gssp add through configuration

MKodde:
This pull request is the SelfService portion of making the generic second factors configurable through YML configuration. To do so, a service was introduced that reads the gssf configuration and helps the application with LOA related queries on a SecondFactorType value object.

This pull request is related to this [story](https://www.pivotaltracker.com/story/show/144698955)


src/Surfnet/StepupSelfService/SamlStepupProviderBundle/Provider/ViewConfig.php
```diff
@@ -0,0 +1,236 @@
...
+
+    /**
+     * @var array
+     */
+    private $popFailed;
+
+    /**
+     * @var null
+     */
+    public $currentLanguage = null;
```
jorissteyn: It's better to solve this problem in the interface instead of having to modify state from outside this object.

Can't we solve this on the DI level? For example, in SurfnetStepupSelfServiceSamlStepupProviderExtension:

```
$viewConfigDefinition = new Definition('Surfnet\StepupSelfService\SamlStepupProviderBundle\Provider\ViewConfig', [
    new Reference('request_stack'),
    $configuration['view_config']['loa'],
    ...
]);
```

MKodde: Yes, good proposal. Fixed this.

PR #126, opened 2017-06-19
Clear:left after every third column of the second factor selection list

MKodde:
Tested in Firefox and Chrome on Linux. On different view port widths.

This pull request fixes the problem when the list of second factors is not ordered in rows from left to right. The second row would start underneath the last item of the first row.

This is a common Bootstrap issue. Bootstrap dictates a row element for each row of columns. This is not always an option. And in that case a css workaround can be used to fix the styling issue.

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/public/less/style.less
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/public/less/style.less
@@ -19,13 +19,22 @@
     }
 }
 
-.second-factor-selector {
-    text-align: center;
+.second-factors {
 
-    img {
-        margin-top: 1em;
-        max-width: 200px;
-        max-height: 90px;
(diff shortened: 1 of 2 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/public/less/style.less
```diff
@@ -19,13 +19,22 @@
...
 
-.second-factor-selector {
-    text-align: center;
+.second-factors {
 
-    img {
-        margin-top: 1em;
-        max-width: 200px;
-        max-height: 90px;
+    @media (min-width: 992px) {
```
jorissteyn: Better to not hard-code those. Use @screen-md!

MKodde: Good point.

PR #138, opened 2018-03-14
Stop assuming Tiqr is the only GSSP

MKodde:
Remove the hard tiqr and biometric references from the translation file and provide a means to translate the token name in the front end.

The solution was based on a very similar problem solved in [RA](https://github.com/OpenConext/Stepup-RA/pull/149).


MKodde:
:warning: Note that the Stepup-bundle version upgrade is only rolled out to SS & RA. As only those two applications benefit from the version bump.

PR #139, opened 2018-03-15
Use the generic error page for all exceptions

jorissteyn:
Instead of rendering specific error pages on a case-by-case basis,
handle all exceptions in the stepup-bundle exception controller. This
way, all error pages look the same with only the page title and error
description tailored to the specific error. Logging of exceptions is
also handled by the exception controller.

The following error templates are removed and the error is handled
using the exception controller:

 * authnFailed.html.twig
 * preconditionNotMet.html.twig
 * blank page in case of 'access denied'

And specific user error messages are shown for the following error
conditions:

 * response is missing a required attribute
 * unknown service provider
 * request not signed
 * request signed with unsupported algorithm
 * request signature validation failed

See:
 - https://www.pivotaltracker.com/story/show/137516239
 - https://www.pivotaltracker.com/story/show/155515714


app/Resources/translations/messages.nl_NL.xliff
```diff
@@ -642,6 +579,139 @@ Er is een e-mail met activatiecode gestuurd naar het e-mailadres %email%. Volg d
...
+        <jms:reference-file line="11">/../vendor/surfnet/stepup-bundle/src/Resources/views/Exception/error404.html.twig</jms:reference-file>
+      </trans-unit>
+      <trans-unit id="45278127b39b1029ff5941470adcab5cc0358c58" resname="stepup.error.error_code">
+        <source>stepup.error.error_code</source>
+        <target>Foutcode</target>
+        <jms:reference-file line="24">/../vendor/surfnet/stepup-bundle/src/Resources/views/Exception/error.html.twig</jms:reference-file>
+      </trans-unit>
+      <trans-unit id="91d524414db2193ad8462909ec5da4c9a732b595" resname="stepup.error.generic_error.description">
+        <source>stepup.error.generic_error.description</source>
+        <target>Something went wrong. Please try again.</target>
```
MKodde: Should be translated?!

src/Surfnet/StepupSelfService/SelfServiceBundle/Security/Authentication/Provider/SamlProvider.php
```diff
@@ -118,7 +119,7 @@ private function getSingleStringValue($attribute, AssertionAdapter $translatedAs
         $values = $translatedAssertion->getAttributeValue($attribute);
 
         if (empty($values)) {
-            throw new BadCredentialsException(sprintf('Missing value for required attribute "%s"', $attribute));
+            throw new MissingRequiredAttributeException(sprintf('Missing value for required attribute "%s"', $attribute));
```
MKodde: When manually triggering this exception, my Request ID is always empty in the error page. Maybe this is wanted behavior is this the case?

From a users perspective, it might make more sense to remove empty values from the details table on the error page.

PR #140, opened 2018-03-28
Show a single token test button

MKodde:
* The test endpoint was changed to no longer test a specified token type.
* To ascertain the user is allowed to use the token test endpoint, a new second factor service method was added. Testing if the user has a vetted LoA >= 2 token.
* The test button is placed in the footer of the vetted tokens table.

Details and a screenshot can be found in:
https://www.pivotaltracker.com/story/show/154476963


jorissteyn:
Tested, looks good but I'm not conviced we need to implement the minimum loa check here. Let's discuss this on Tuesday.

MKodde:
The check has been simplified (scrapped the >= LoA 2 check). The check is now only on vetted second factors for a given entity.

PR #142, opened 2018-04-03
Improve token selection ordering

MKodde:
The sort order of the available tokens was changed. To achieve this  goal, a collection object was added to the project collecting the GSSP and hard coded tokens.

Story details can be found in Pivotal
https://www.pivotaltracker.com/story/show/156311244

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Value/AvailableTokenCollection.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Value/AvailableTokenCollection.php
@@ -30,16 +30,16 @@ class AvailableTokenCollection
     private $collection = [];
 
     /**
-     * @param array $hardcodedTokens
+     * @param array $builtInTokens
      * @param array $gsspTokens
      * @return AvailableTokenCollection
      */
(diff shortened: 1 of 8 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Value/AvailableTokenCollection.php
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    private function sortCollection()
+    {
+        // The collection is first sorted by LoA level and then in alphabetic order.
+        uasort($this->collection, function (AvailableTokenInterface $a, AvailableTokenInterface $b) {
+            if ($a->getLoaLevel() === $b->getLoaLevel()) {
+                return strcmp($a->getType(), $b->getType());
+            }
+            return $a->getLoaLevel() > $b->getLoaLevel();
```
jorissteyn: Should not be a boolean! -1/0/1

MKodde: Thanks for the feedback. As discussed in IRL, I've also renamed the HardcodedToken to BuiltIn token.

PR #148, opened 2018-05-25
Redirect after POST when returning from GSSP

jorissteyn:
The ACS location of GSSP should never show a page without first
redirecting. This way, the user can refresh the page or change the
language before trying again.

See: https://www.pivotaltracker.com/story/show/157843953

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/Registration/GssfController.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/Registration/GssfController.php
@@ -35,6 +35,17 @@ use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 final class GssfController extends Controller
 {
     /**
+     * Render the initiation form.
+     *
+     * This action has two parameters:
+     *
+     * - authenticationFailed (default false), will trigger an error message
+     *   and is used when a SAML failure response was received, for example
+     *   when the users cancelled the registration
+     *
+     * - proofOfPossessionFailed (default false), will trigger an error message
+     *   when possession was not proven, but the SAML response was successful
+     *
      * @param Request $request
      * @param string $provider
      * @return array|Response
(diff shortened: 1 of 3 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/Registration/GssfController.php
```diff
@@ -105,8 +113,8 @@ public function consumeAssertionAction(Request $httpRequest, $provider)
                 sprintf('Could not process received Response, error: "%s"', $exception->getMessage())
             );
 
-            return $this->renderInitiateForm(
-                $provider->getName(),
+            return $this->redirectToInitiationForm(
+                $provider,
                 ['authenticationFailed' => true]
```
MKodde: Why pass the provider alongside the options when all you do in the `redirectToInitiationForm` is: adding the provider name to the options?

src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/config/routing.yml
```diff
@@ -63,7 +63,10 @@ ss_registration_sms_prove_possession:
 ss_registration_gssf_initiate:
     path:     /registration/gssf/{provider}/initiate
     methods:  [GET]
-    defaults: { _controller: SurfnetStepupSelfServiceSelfServiceBundle:Registration/Gssf:initiate }
+    defaults:
+        _controller: SurfnetStepupSelfServiceSelfServiceBundle:Registration/Gssf:initiate
+        authenticationFailed: false
```
MKodde: Maybe add a doc block for the initiate action in the `GssfController` explaining the defaults that are set for the action and what their intention is.

jorissteyn: I've documented them in the controller action docblock.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/Registration/GssfController.php
jorissteyn: We could pass a string, but then I can't be sure the string refers to a valid GSSP provider.

PR #155, opened 2018-09-13
Prevent form loading issue

MKodde:
The feature in Symfony 2 where forms where created by referencing them
by their name (ss_verify_sms_challenge) no longer works in Symfony 3.
This commit fixes this by referencing the form by it's FQDN
(VerifySmsChallengeType).

In the previous Symfony 3.4 compatibility PR, a lot of these fixes have
been done for other Types. This one was overlooked.

See: https://www.pivotaltracker.com/story/show/160473876


MKodde:
I've just amended two more fixes of the same type onto the existing commit.

MKodde:
Like on my RA PR, merged this to `develop` and backported the fix to the `2.10` release branch. Any successive bugfixes will be fixed in a similar manner.

PR #158, opened 2018-10-16
Support the amended institution parameter

MKodde:
The VerifiedSecondFactorSearchQuery now takes the institution as
one of its arguments. This argument is mandatory and is now added
to all the calls of the VerifiedSecondFactorSearch service call.

See https://www.pivotaltracker.com/story/show/160283514

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SecondFactorService.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SecondFactorService.php
@@ -253,30 +253,6 @@ class SecondFactorService
     }
 
     /**
-     * @param string $secondFactorId
-     * @param string $identityId
-     * @return null|string
-     */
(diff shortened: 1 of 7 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SecondFactorService.php
```diff
@@ -175,13 +176,15 @@ public function findUnverifiedByIdentity($identityId)
      * Returns the given registrant's verified second factors.
      *
      * @param string $identityId
+     * @param string $insitution
      * @return VerifiedSecondFactorCollection
      */
-    public function findVerifiedByIdentity($identityId)
+    public function findVerifiedByIdentity($identityId, $insitution)
```
pablothedude: typo?

MKodde: It never happened ;)

PR #162, opened 2019-01-09
Fix mpdf relative url's and logging

pablothedude:
Relative url's weren't supported, because mpdf was trying to get it by
hostname in the request. If however a proxy should update the hostname
it wouldn't work. Therefore the `base` html tag was added which was
supported by mpdf for resolving absolute urls. Also the
logging was added and decent symfony shutdown.

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RegistrationController.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RegistrationController.php
@@ -26,6 +26,7 @@ use Surfnet\StepupSelfService\SelfServiceBundle\Service\SecondFactorService;
 use Surfnet\StepupSelfService\SelfServiceBundle\Value\AvailableTokenCollection;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\ResponseHeaderBag;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
 class RegistrationController extends Controller
@@ -165,8 +166,6 @@ class RegistrationController extends Controller
     /**
      * @param $secondFactorId
      * @return Response
-     *
-     * @SuppressWarnings(PHPMD.ExitExpression) MPDF requires bypassing Symfony, so we exit() when MPDF is done.
      */
     public function registrationPdfAction($secondFactorId)
     {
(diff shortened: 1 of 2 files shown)


pablothedude:
https://www.pivotaltracker.com/story/show/161719264

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RegistrationController.php
```diff
@@ -178,9 +179,25 @@ public function registrationPdfAction($secondFactorId)
                 'tempDir' => sys_get_temp_dir(),
             )
         );
+        $mpdf->setLogger($this->get('logger'));
```
MKodde: If we can use the solution you propose than please remove the suppression that was previously added.

>  @SuppressWarnings(PHPMD.ExitExpression) MPDF requires bypassing Symfony, so we exit() when MPDF is done.

See #134 for the original development effort.

I did not see any requests for additional logging from mpdf. Personally I think it's a good idea, did you run this past @pmeulen or another product owner?

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RegistrationController.php
```diff
@@ -178,9 +179,25 @@ public function registrationPdfAction($secondFactorId)
...
+        $response->headers->set('Content-Description', 'File Transfer');
+        $response->headers->set('Content-Transfer-Encoding', 'binary');
+        $response->headers->set('Cache-Control', 'public, must-revalidate, max-age=0');
+        $response->headers->set('Pragma', 'public');
+        $response->headers->set('Expires', 'Sat, 26 Jul 1997 05:00:00 GMT');
+        $response->headers->set('Last-Modified', '' . gmdate('D, d M Y H:i:s') . ' GMT');
+        $response->headers->set('Content-Type', 'application/pdf');
 
-        exit;
+        return $response;
```
MKodde: The PDF is not presented to the user as a download but opens in the PDF viewer of the browser.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RegistrationController.php
pablothedude: Yes, this issue wasn't logged either.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RegistrationController.php
pablothedude: I had to set the disposition header myself. I've added this so now it should work as expected.

PR #163, opened 2019-01-10
Fix art for authentication

pablothedude:
This will fix the art for exceptions when authenticating. The art
was based on the users UUID but should be generated per exception type
and should not be unique per user.


MKodde:
Some background information for future readers. The ART is based on the exception message test. Any text between quotes or single quotes is removed as they usually contain variable text. We want an ART code that is the same between every time that specific error is logged. 

In this instance, the variable text was parsed in square brackets. And that caused the ART code to be different every time the exception was raised.

@pablothedude this fixes the issue at hand. As discussed in the Pivotal story. We should think about making a more robust solution for this. Or improving the docs to help prevent this kind of issues in the future.

PR #165, opened 2019-01-10
Open help in new tab

pablothedude:
The help link in the footer should op en in a new tab.


pablothedude:
https://www.pivotaltracker.com/story/show/161718649

PR #167, opened 2019-01-21
Perform security related chores

MKodde:
[ ] Install and use the Security Checker
[x] Travis used a custom set of composer dependencies. That has been
undone now that the middleware client bundle version 3 is released.

See: https://www.pivotaltracker.com/story/show/163226922


.travis.yml
```diff
@@ -21,11 +21,6 @@ cache:
 
 before_script:
   - phpenv config-add .travis.php.ini
-  # TODO: cleanup after FGA implementation!
```
pablothedude: :+1:

PR #168, opened 2019-01-28
Create a pdf base template to support local url's

pablothedude:
In order to be able to view pdf's in production a template with only
local urls' had to be created.

https://www.pivotaltracker.com/story/show/161719264


pablothedude:
The problem was the reuse of the main site template for both pdf and html. Therefore a new base template had to be created in order to prevent loading the logo with a request. This had nothing to do with base64 encoding the logo else it should be also encoded for html pages and this is far from optimal in my opinion.
The initial problem was that because the host header was used by the mpdf library to convert relative url's to absolute url's it tried to fetch the image directly from load balancer by ip-adres. So the initial fix was to set the base url so mpdf would resolve absolute url's not with the hostname header of the request but with the html base tag. And that seemed to worked as expected only on the test environment the hostname of selfservice could not be resolved which should be fixed now by splitting the base templates.

PR #171, opened 2019-05-02
Find verified second factors without authorization context

MKodde:
The authorization context that was added for the RA application applies
an authorization context to the query. This filters out results that not
match the context of the current RA state (the institution that is
currently active).

For Self Service we do not want to apply that context based filtering.

See: https://www.pivotaltracker.com/story/show/165740432
See: https://github.com/OpenConext/Stepup-Middleware/pull/279
See: https://github.com/OpenConext/Stepup-Middleware-clientbundle/pull/82

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SecondFactorService.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SecondFactorService.php
@@ -19,7 +19,7 @@
 namespace Surfnet\StepupSelfService\SelfServiceBundle\Service;
 
 use Surfnet\StepupMiddlewareClient\Identity\Dto\UnverifiedSecondFactorSearchQuery;
-use Surfnet\StepupMiddlewareClient\Identity\Dto\VerifiedSecondFactorSearchQuery;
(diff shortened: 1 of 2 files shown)


MKodde:
After MW and MW client bundle PRs have been merged, we can install the MW client bundle release into this PR in order to actually use the new endpoint.

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/SecondFactorService.php
```diff
@@ -189,7 +189,7 @@ public function findVerifiedByIdentity($identityId, $actorInstitution)
         $query->setInstitution($actorInstitution);
         // Actor and identity are equal in SelfService.
         $query->setActorId($identityId);
-        return $this->secondFactors->searchVerified($query);
+        return $this->secondFactors->searchVerifiedWithoutAuthorizationContext($query);
```
pablothedude: Please rename to something like `searchOwn` or something like that.

pablothedude: LGTM!

PR #172, opened 2019-05-09
Remove the actorInstitution

pablothedude:
The actorInstitution was removed. this is possible because the SRAA (institution) switcher is removed as it is no longer useful after the FGA changes. This because earlier all overviews where from the viewpoint of an institution but now it's from a registration authority. 

These changes are needed to reflect the changes in RA and Middleware.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,7 +1,7 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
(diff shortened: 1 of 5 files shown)


pablothedude:
https://www.pivotaltracker.com/story/show/165759964

pablothedude:
I've squashed the `docblock` commit in the previous.

PR #175, opened 2019-11-07
Upgrade xmlseclibs to version 3.0.4

MKodde:
This change will apply the countermeasures to harden against CVE 2019-3465 and will effectively bump `robrichards/xmlseclibs` to version 3.0.4

Targeted at release 16 (2.10.7)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,13 @@ before_script:
 script:
   - ant
 
-
 branches:
   only:
     - master
     - develop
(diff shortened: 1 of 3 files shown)


MKodde:
Yes, the more I think about it, the more I agree with you and @pmeulen 
No sense in taking unnesecary risk in also updating other packages.

I've reconstructed my commit and now only upgrade the xmlseclibs

PR #177, opened 2020-01-06
Install various security updates

MKodde:
Install recent Symfony security updates

Also pins twig/twig to a compatible version for use with the Mopa Bootstrap bundle.


src/Surfnet/StepupSelfService/SamlStepupProviderBundle/DependencyInjection/Compiler/ViewConfigCollectionPass.php
```diff
@@ -43,6 +43,7 @@ public function process(ContainerBuilder $container)
         $taggedServices = array_keys($taggedServices);
 
         foreach ($taggedServices as $id) {
+            $gsspIdMatches = null;
```
pablothedude: I know this is possible but should be an empty array. Feel free to remove or leave as-is.

PR #181, opened 2020-02-11
Provide identity encryption functionality

MKodde:
The IdentityEncrypter is to be used to encrypt and write the attributes
received from one of the Remote vetting services (readid, idin or irma).
These attributes are encrypted with openssl public key encryption (can
be decrypted using the associated private key).

The important parts of this code is covered in test, an integration test
was written to prove the correctness of the encrypter.

No SAML response assertion is available yet from one of the services,
but a DTO is created that is to be used to contain these attributes. For
now bogus data is used to test the encrypter.

https://www.pivotaltracker.com/story/show/166932716


app/config/remote_vetting.yml
```diff
@@ -10,8 +11,21 @@ parameters:
     assertionConsumerUrl: https://selfservice.stepup.example.com/second-factor/acs
     privateKey: "%saml_sp_privatekey%"
 
+  identity_encryption_configuration:
+    encryption_public_key: |-
```
pablothedude: Could we get the key through a file. I would love to use the key myself for testing purposes.

app/config/remote_vetting.yml
```diff
@@ -32,25 +52,21 @@ services:
       $logger: "@logger"
 
   Surfnet\StepupSelfService\SelfServiceBundle\Controller\RemoteVettingController:
-    class: Surfnet\StepupSelfService\SelfServiceBundle\Controller\RemoteVettingController
     arguments:
       - "@self_service.remote_vetting.service"
 
+  Surfnet\StepupSelfService\SelfServiceBundle\Mock\RemoteVetting\MockGateway:
```
pablothedude: Is this change needed? Seems to be moved only a few lines up?

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RemoteVetting/RemoteVettingController.php
```diff
@@ -0,0 +1,142 @@
+<?php
```
pablothedude: Is this file not also one level down?

src/Surfnet/StepupSelfService/SelfServiceBundle/RemoteVetting/Configuration/RemoteVettingConfiguration.php
```diff
@@ -0,0 +1,54 @@
+<?php
```
pablothedude: Why is RemoteVetting in a separate folder, should the Locale logic then also be moved to a Locale folder? Would you elaborate that the reason was to do so? Because I doubt if someone is making changes to the codebase in a few years from now if the separation is so clear?

src/Surfnet/StepupSelfService/SelfServiceBundle/RemoteVetting/Service/IdentityEncrypter.php
```diff
@@ -0,0 +1,65 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\StepupSelfService\SelfServiceBundle\RemoteVetting\Service;
+
+use Surfnet\StepupSelfService\SelfServiceBundle\RemoteVetting\Configuration\RemoteVettingConfiguration;
+use Surfnet\StepupSelfService\SelfServiceBundle\RemoteVetting\Dto\IdentityDto;
+
+class IdentityEncrypter
```
pablothedude: Nice separation!

src/Surfnet/StepupSelfService/SelfServiceBundle/RemoteVetting/Service/IdentityFilesystemWriter.php
```diff
@@ -0,0 +1,51 @@
...
+     */
+    public function write($data)
+    {
+        $fileName = $this->createFileName($this->configuration->getLocation());
+        file_put_contents($fileName, $data);
+    }
+
+    private function createFileName($location)
+    {
+        $fileName = uniqid('identity_', true);
```
pablothedude: Could we not inject an id, for example a RequestId or something like that so the attributes could be more easily traced back?

src/Surfnet/StepupSelfService/SelfServiceBundle/Tests/RemoteVetting/Dto/AttributeTest.php
```diff
@@ -0,0 +1,68 @@
...
+        yield ['foobar', ['foobar', 'foobar2']];
+        yield ['foobar', ['foobar', 'foobar2', 'foobar3']];
+        yield ['foobar', ['foobar', 2]];
+        yield ['foobar', [true, 2]];
+        yield ['foobar', [true]];
+    }
+
+    public function provideInvalidAttributes()
+    {
+        yield ['foobar', [['foobar']], 'The $value of an Attribute must be a scalar or array value'];
```
pablothedude: This looks like a scalar and array value...
There are some other occurrences as well, it looked a bit strange, but feel to leave it as-is.

src/Surfnet/StepupSelfService/SelfServiceBundle/Tests/RemoteVetting/Service/IdentityEncrypterIntegrationTest.php
```diff
@@ -0,0 +1,170 @@
...
+ */
+class IdentityEncrypterIntegrationTest extends UnitTest
+{
+    private $encrypter;
+
+    private $config;
+
+    private $writer;
+
+    private $publicKey = <<<CERT
```
pablothedude: Could we use files for the keys?

src/Surfnet/StepupSelfService/SelfServiceBundle/RemoteVetting/Configuration/RemoteVettingConfiguration.php
MKodde: I see this as a collection of cohesive application parts (all related to RemoteVetting) collected in a namespace of their own. I find it convenient to have those classes together.

src/Surfnet/StepupSelfService/SelfServiceBundle/Tests/RemoteVetting/Dto/AttributeTest.php
MKodde: The attribute value can be an array of scalar values as per the saml spec. That's whats being tested here.

src/Surfnet/StepupSelfService/SelfServiceBundle/Tests/RemoteVetting/Service/IdentityEncrypterIntegrationTest.php
MKodde: We sure could, it might be cleaner from a code perspective.

PR #187, opened 2020-02-25
Fix RV acceptance tests in Travis

pablothedude:
Disable garbage collection for acceptance test

For the acceptance tests the garbage collection is disabled in
order to prevent a segfaults in Travis. This should be fixed in php 7.

See: https://bugs.php.net/bug.php?id=72286

### DIFF
--- a/app/config/remote_vetting.yml
+++ b/app/config/remote_vetting.yml
@@ -83,12 +83,14 @@ services:
     $logger: '@logger'
 
   Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVettingService:
+    public: true
     arguments:
       $remoteVettingContext: '@Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVetting\RemoteVettingContext'
       $identityEncrypter: '@Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVetting\Encryption\IdentityEncrypter'
       $logger: '@logger'
 
   Surfnet\StepupSelfService\SelfServiceBundle\Controller\RemoteVettingController:
+    public: true
     arguments:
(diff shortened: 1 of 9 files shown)


app/config/remote_vetting.yml
```diff
@@ -76,19 +76,21 @@ services:
...
-      $remoteVettingService: '@Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVettingService'
-      $logger: '@logger'
+    $identityProviderFactory: '@Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVetting\IdentityProviderFactory'
+    $serviceProviderFactory: '@Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVetting\ServiceProviderFactory'
+    $postBinding: '@surfnet_saml.http.post_binding'
+    $remoteVettingService: '@Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVettingService'
+    $logger: '@logger'
 
   Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVettingService:
+    public: true
```
MKodde: Making the services public seems unnecessary. I found one instance where you loaded RVService directly from the container (in web test). But the Controllers do not seem to require public access.

Also I've been looking at this config file lately, and I think we need to move the parameters to the regular parameters.yml. This to keep the application deployable without too many additional build steps.

pablothedude: The `public` was done in order to fix deprecation warnings.

The parameters should indeed be removed in the end but during development I would like to keep it this way because RV config is in one place and in our dev setup configs are in VCS and therefore easily changed.

MKodde: Thanks for the reply! Makes perfect sense. I'll add a task/ticket to Pivotal to ensure we do not forget to move the parameters at the end.

PR #188, opened 2020-02-28
Add data mapper for form attribute matches

pablothedude:
The attrbute matches needed to be mapped with DI.
This is done with the help of a datamapper.

Please review #187 first.
Be aware that the base branch has to be changed before merging.

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Command/RemoteVetValidationCommand.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Command/RemoteVetValidationCommand.php
@@ -18,21 +18,12 @@
 
 namespace Surfnet\StepupSelfService\SelfServiceBundle\Command;
 
-use Surfnet\StepupMiddlewareClientBundle\Identity\Dto\Identity;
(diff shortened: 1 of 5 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Command/RemoteVetValidationCommand.php
```diff
@@ -18,21 +18,12 @@
...
 class RemoteVetValidationCommand
 {
     /**
-     * The attributes and their state
-     *
-     *  [key => [
-     *      'valid' => true,
-     *      'remarks' => 'remarks',
-     *  ]]
+     * The attribute matches
```
MKodde: This is not that useful to me as a reader it's not entirely clear what the purpose of this field is. My suggestion is to either remove the doc block, or add additional explanation.

pablothedude: Removed it.

pablothedude:
Updated the description

PR #191, opened 2020-03-12
Finalize identity data encryption logic

MKodde:
Summary of work done: 

1. The attributes have been aggregated
2. The other metedata stored alongside the attributes is collected and set on a identity data VO
3. Tied together these new concepts
4. Tried to add relevant tests

Read commit messages for details.

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Service/RemoteVetting/Dto/AttributeListDto.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Service/RemoteVetting/Dto/AttributeListDto.php
@@ -19,7 +19,10 @@
 namespace Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVetting\Dto;
 
 use phpDocumentor\Reflection\Types\Self_;
+use SAML2\XML\saml\NameID;
 use Serializable;
+use Surfnet\SamlBundle\SAML2\Attribute\Attribute;
+use Surfnet\SamlBundle\SAML2\Attribute\AttributeSet;
 use Surfnet\StepupSelfService\SelfServiceBundle\Assert;
 use Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVetting\Value\Attribute;
 use Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVetting\Value\AttributeCollection;
@@ -50,6 +53,36 @@ class AttributeListDto implements Serializable, AttributeCollectionInterface
 
         $this->attributes = new AttributeCollection($attributes);
         $this->nameId = $nameId;
+    }
+
+    /**
+     * @param AttributeSet $attributeSet
+     * @return AttributeListDto
+     */
+    public static function fromAttributeSet(AttributeSet $attributeSet)
+    {
+        $attributes = [];
+        $nameID = '';
+        /** @var Attribute $attribute */
+        foreach ($attributeSet as $attribute) {
+            $name = $attribute->getAttributeDefinition()->getName();
+            $values = $attribute->getValue();
+            foreach ($values as $value) {
+                if ($value instanceof NameID) {
+                    $nameID = (string)$value->value;
+                    continue;
+                }
+
+                // Singular values wrapped in an array are extracted for convenience
+                if (is_array($values) && count($values) === 1) {
+                    $values = reset($values);
+                }
+
+                $attributes[$name] = $values;
+            }
+        }
+
+        return new self($attributes, $nameID);
     }
 
     /**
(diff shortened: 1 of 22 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Service/RemoteVetting/Value/AttributeCollection.php
```diff
@@ -45,9 +45,17 @@ private function add(Attribute $attribute)
...
     public function getIterator()
     {
         return new ArrayIterator($this->attributes);
     }
+
+    public function getAttributes()
+    {
+        $output = [];
+        foreach ($this->attributes as $attribute) {
+            $output[$attribute->getName()] = $attribute->jsonSerialize();
```
pablothedude: Why is serialization needed in a collection?

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/RemoteVetting/Value/AttributeMatchCollection.php
```diff
@@ -69,4 +69,13 @@ public function offsetGet($offset)
     {
         return isset($this->matches[$offset]) ? $this->matches[$offset] : null;
     }
+
+    public function getAttributes()
```
pablothedude: Please remove duplicate code. See remark above.

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/RemoteVetting/Dto/AttributeListDto.php
```diff
@@ -18,17 +18,20 @@
 
 namespace Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVetting\Dto;
 
+use phpDocumentor\Reflection\Types\Self_;
```
pablothedude: Was this intentionally?

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/RemoteVetting/Value/AttributeCollection.php
MKodde: Every collection, that implements the `AttributeCollectionInterface` should be able to do so. This way every implementation is responsible for how their data would be represented when displayed in the encrypted file. 

> Why is serialization needed in a collection?
The data that ends up in the encrypted file needs to be serializable, mainly to make things readable and structured in a recognizable manner. The collection is easily held responsible for this.

> Couldn't this be better moved to the IdentityData class itself with a dedicated serialization method? That way the collection is only responsible for the collection? Or even better a dedicated serializer.

That would be a possibility. But I judge against that. By utilizing the composite pattern this goal can be achieved in a more maintenance-friendly manner. Also, creating a dedicated serializer for this would create a very messy class that would constantly type check a collection and perform slightly different serialization logic for each collection.

MKodde: I did find that the `jsonEncode` call on the attribute should be removed. This will result in double encoding.

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/RemoteVetting/Value/AttributeMatchCollection.php
MKodde: Also, see my response to your feedback. What should be changed in this method is the call to `jsonEncode` on the `$match`

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/RemoteVetting/Dto/AttributeListDto.php
MKodde: Good catch. No that was added by mistake. Also removed the stale `AttributeMatchCollection`

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/RemoteVetting/Value/AttributeCollection.php
pablothedude: If the serialization would be only on the IdentityData object it would be more maintenance friendly because then the code would be in one place. Certainly when the serialization is only needed for that object. The serialization could also be removed from the AttributeListDto object because it won't be needed there. That way double encoding would also be prevented. Maybe we could decide in the end if we want to refactor that. Thanks for elaborating!
Could you if you keep it this way change the name of the interface to for example `AttributeCollectionSerializationInterface` and the method to something like `getSerializedAttributes`.
Feel free to use an alternative the main reason is to clarify it's usage.

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/RemoteVetting/Value/AttributeMatchCollection.php
pablothedude: We could address this later on as suggested.

PR #196, opened 2020-03-18
Move IdentityAggregation to Remote vetting service

pablothedude:
Move the aggregation of the IdentityData to the RemoteVettingService.
That way the controller will keep lightweight and the responsibility
of remote vetting is delegated and only glue is used in the
controller. This resulted in less coupling.

Also the ProviderName is set in the IdentityData, this needed to be
done after the feature was implemented on a different branch.

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RemoteVettingController.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RemoteVettingController.php
@@ -29,10 +29,8 @@ use Surfnet\StepupSelfService\SelfServiceBundle\Security\Authentication\Token\Sa
 use Surfnet\StepupSelfService\SelfServiceBundle\Service\ApplicationHelper;
 use Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVetting\Dto\AttributeListDto;
 use Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVetting\Dto\RemoteVettingTokenDto;
-use Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVetting\Encryption\IdentityData;
 use Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVetting\RemoteVettingViewHelper;
 use Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVetting\SamlCalloutHelper;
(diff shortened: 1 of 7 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Service/RemoteVettingService.php
```diff
@@ -112,15 +127,48 @@ public function done(ProcessId $processId, IdentityData $identityData)
...
+        Identity $identity,
+        AttributeListDto $localAttributes,
+        AttributeMatchCollection $attributeMatches,
+        $remarks
+    ) {
+        $nameId = $identity->nameId;
+        $institution = $identity->institution;
+        $version = $this->applicationHelper->getApplicationVersion();
+        $remarks = (string)$remarks;
+        $remoteVettingSource = $this->remoteVettingContext->getIdentityProviderName();
```
MKodde: Hurray!

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RemoteVettingController.php
```diff
@@ -210,46 +202,28 @@ public function remoteVetMatchAction(Request $request, $processId)
         /** @var SamlToken $samlToken */
         $samlToken = $this->container->get('security.token_storage')->getToken();
 
-        $attributes = AttributeListDto::fromAttributeSet($samlToken->getAttribute(SamlToken::ATTRIBUTE_SET));
+        $localAttributes = AttributeListDto::fromAttributeSet($samlToken->getAttribute(SamlToken::ATTRIBUTE_SET));
```
MKodde: Consider naming this variable identity-provider-attributes as I later refer to them with that moniker when they get encrypted.

pablothedude: I've added a commit and renamed the 'identity-provider-attributes' to 'local-attributes'. Because local and remote are both IdP's.

pablothedude:
The reasoning behind this is that `remote` and `local` is more clarifying when you're dealing with remote vetting and will leave out a lot of guessing. I don't believe that a `identity-provider-attributes` will be more clarifying because you don't have a single IdP in remote vetting and I reckoned is therefore less distinctive.

Maybe it would be an idea to document the used ubiquitous language? But for the POC I thinkk this will be overkill. I'll create an issue with things we would like when we'll come out of the POC phase.

PR #199, opened 2020-03-19
Remove remote veting development efforts

pablothedude:
Remove the remote vetting development efforts before releasing an MVP.

https://www.pivotaltracker.com/story/show/171571539

PR #200, opened 2020-03-20
Prevent remote vetting of expired tokens

pablothedude:
Only non expired tokens could be remote vetted. This wasn't fixed
earlier because it came in handy during development.

https://www.pivotaltracker.com/story/show/171571641
https://www.pivotaltracker.com/story/show/171571556

Please review #199 first

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RemoteVettingController.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RemoteVettingController.php
@@ -21,6 +21,7 @@ use Exception;
 use Psr\Log\LoggerInterface;
 use SAML2\Response\Exception\PreconditionNotMetException;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
+use Surfnet\StepupBundle\DateTime\RegistrationExpirationHelper;
 use Surfnet\StepupSelfService\SelfServiceBundle\Command\RemoteVetCommand;
 use Surfnet\StepupSelfService\SelfServiceBundle\Command\RemoteVetValidationCommand;
 use Surfnet\StepupSelfService\SelfServiceBundle\Exception\InvalidRemoteVettingContextException;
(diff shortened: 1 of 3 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RemoteVettingController.php
```diff
@@ -93,14 +100,12 @@ public function remoteVetAction(Request $request, $secondFactorId, $identityProv
         }
 
         $secondFactor = $service->findOneVerified($secondFactorId);
-        if ($secondFactor === null) {
+        if ($secondFactor === null || $this->expirationHelper->hasExpired($secondFactor->registrationRequestedAt)) {
```
MKodde: Good call to use the expirationHelper here! Would it not be a good defense in depth measure to also verify the token has not expired in the `remoteVetMatchAction`? Highly unlikely the token expired in the meantime, but it is certainly possible.

pablothedude: I deliberately didn't add a check on the match page to be user friendly. If a RV process is started in time then why not let it finish. Then it would be better to let a user not even start in the first place. The matching itself could now possibly take long but that shouldn't be happening when the final algorithm is implemented.

I certainly agree we have to make a informed decision about this. I'll add your concerns to the things to do after MVP so we keep this in scope and could let our PM decide what should be guarded.

Be aware that the time of expiration doesn't seem to be so clear either:
https://www.pivotaltracker.com/story/show/90087146

pablothedude:
Added the Pivotal issue to the original description.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/RemoteVettingController.php
MKodde: I'm fine with implementing it at a later stage, but I do think that when you somehow drop into the second portion of the verification, you should also have this token expired check. To be clear, in __addition__ to the check you already implemented!

But as long as we have this covered on the backlog im all :+1:

PR #201, opened 2020-03-20
Add complete remote vetting E2E test

pablothedude:
Add a remote vetting E2E test to validate the flow
of the remote vetting process.

https://www.pivotaltracker.com/story/show/171571414

Please review #200 first


src/Surfnet/StepupSelfService/SelfServiceBundle/Tests/Mock/RemoteVetting/MockRemoteVetControllerTest.php
```diff
@@ -174,14 +189,76 @@ public function an_unsuccessful_response_from_a_remote_vetting_idp_should_fail()
...
         $this->assertEquals('https://selfservice.stepup.example.com/overview', $this->client->getRequest()->getUri());
         $this->assertContains('Unable to validate the information', $this->client->getResponse()->getContent());
     }
 
+
+    /**
+     * @test
+     * @group rv
+     */
+    public function a_verified_token_must_be_vetted_with_external_idp_e2e()
```
MKodde: Is this single use case enough coverage? We might want to cover some other aspects? Like verify things are handled correctly when things go wrong. Or the ability for the user to cancel out of a RV action

pablothedude: This is indeed not fully tested but only partially when a  response is returned from the IdP.
See for example:
`an_unsuccessful_response_from_a_remote_vetting_idp_should_fail`
`a_user_cancelled_response_from_a_remote_vetting_idp_should_fail`

I think that will be sufficient for now but  maybe we want to have more extensive tests when the first test results of the RV IdP's are being processed. This is also dependent on how an  IdP implements for example how a user cancels and what we receive as ACS response.

PR #202, opened 2020-03-20
Move remote vetting parameters to dist file

pablothedude:
The remote vetting parameters need to be moved to be independent
of install. This was not done during development to keep
all parameters the same during development.

https://www.pivotaltracker.com/story/show/171928781

Please review #201  first


pablothedude:
Thanks, will keep your commit, it's a valuable commit.
Also created an issue and updated the original description accordingly.

PR #204, opened 2020-07-15
Really remove Mopa

pablothedude:
It seems that there still was a reference to Mopa through the Stepup-Bundle.


pablothedude:
This seems to do with Mopa not being complient with newer js version, will close this because an upgrade to Encore seems a better choice.

PR #208, opened 2020-11-06
Upgrade QA tooling and move to PHP72

mroest:
In preparation to the SF4 upgrade this is the first step. This
updates the QA tooling using the best/common practices from the other
Stepup projects.


composer.json
```diff
@@ -13,40 +13,66 @@
...
-    "twig/extensions": "^1.5",
-    "symfony/assetic-bundle": "~2.3",
-    "symfony/monolog-bundle": "^3.1.0",
-    "sensio/distribution-bundle": "^5.0",
-    "sensio/framework-extra-bundle": "~3.0",
-    "incenteev/composer-parameter-handler": "~2.0",
-    "nelmio/security-bundle": "~1.4",
-    "mopa/bootstrap-bundle": "^3.2",
-    "twbs/bootstrap": "^3.3.0",
+    "php": "~7.2",
```
MKodde, line 16: :metal:

.travis.yml
```diff
@@ -8,8 +8,7 @@
 language: php
 
 php:
-  - 5.6
-  - 7.0
+  - 7.2
```
MKodde, line 11: Consider adding the current PHP version (7.4) to see if we start introducing incompatibilities with that version during development.

composer.json
```diff
@@ -13,40 +13,66 @@
...
   },
   "scripts": {
+    "test": [
+      "@lint",
+      "@phpmd",
+      "@phpcs",
+      "@phpcpd",
+      "@phpunit",
+      "@behat",
+      "@security-tests"
```
MKodde, line 59: While at it, why not add the docheader qa check? See stepup-tiqr for an implementation example.

.travis.yml
```diff
@@ -21,18 +20,17 @@ cache:
...
 
 branches:
   only:
     - master
     - develop
-
-addons:
-  apt:
-    packages:
-      - ant
```
MKodde, line 38: :heart:

MKodde (changes requested):
Aside from the docheader question, looks good to go!

.travis.yml
mroest: Yes, but lets do that in a different PR.

composer.json
mroest: Sure, but lets do it in a different PR. I do not want to include too much unrelated stuff and only stick to the php7.2 upgrade

MKodde (approved):

PR #209, opened 2020-11-09
Move to webpack encore

mroest:


composer.json
```diff
@@ -13,40 +13,73 @@
...
-    "mopa/bootstrap-bundle": "^3.2",
-    "twbs/bootstrap": "^3.3.0",
+    "php": "~7.2",
     "fortawesome/font-awesome": "~4.2.0",
-    "jms/translation-bundle": "~1.3.0",
     "guzzlehttp/guzzle": "^6",
+    "incenteev/composer-parameter-handler": "~2.0",
+    "jms/translation-bundle": "~1.3.0",
+    "mopa/bootstrap-bundle": "^3.2",
+    "mopa/composer-bridge": "~1.5",
```
MKodde, line 22: I recon these can be removed.

app/config/webpack_encore.yaml
```diff
@@ -0,0 +1,15 @@
...
+  # If multiple builds are defined (as shown below), you can disable the default build:
+  # output_path: false
+
+  # if using Encore.enableIntegrityHashes() specify the crossorigin attribute value (default: false, or use 'anonymous' or 'use-credentials')
+  # crossorigin: 'anonymous'
+
+  # Cache the entrypoints.json (rebuild Symfony's cache when entrypoints.json changes).
+  # To enable caching for the production environment, creating a webpack_encore.yaml in the config/packages/prod directory with this value set to true
+  # Available in version 1.2
+  #cache: false
```
MKodde, line 14: On Stepup-RA (prod specific config), the cache is enabled for webpack_encore. Consider doing that here to by adding a config/packates/prod/webpack_encore.yaml and override the cache setting there.

MKodde (changes requested):
Hi there, a good starting point here. I compared this PR with its Stepup-RA counterpart. And found some things you might want to address. I see this PR is created prior to the SF4 upgrade. Know that you'll have to change some settings you introduced here when time comes to upgrade to Symfony 4.

1. You will need to remove [this line](https://github.com/OpenConext/Stepup-SelfService/blob/202083fa0f934ca1815cacdee34c41cb515826d5/app/AppKernel.php#L34). I think you will have a 500 error otherwise, as the Mopa Boostrap Bundle is removed.
2. Consider using the framework.asstes.json_manifest_path option. To generate a manifest.json file which can be used in prod mode for cache busting purposes.

I compared your changeset to: `https://github.com/OpenConext/Stepup-RA/pull/237/commits/dd34526c49c723182cad6898e201f39de02561c1`

app/config/config.yml
```diff
@@ -32,29 +33,14 @@ framework:
 
 # Twig Configuration
 twig:
+    form_themes: ['bootstrap_4_layout.html.twig']
```
MKodde, line 36: The magic line ;)

MKodde (changes requested):
Almost there:

1. Please remove https://github.com/OpenConext/Stepup-SelfService/blob/73512d12eac51e00c1b148b5650a9359ecb14224/README.md#L21 and maybe update that section with how the new encore webpack setup works.
2. `https://github.com/OpenConext/Stepup-SelfService/blob/249e9a924d2243d751e1b5a5128297e7dcd5f5c8/src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/views/Form/fields.html.twig#L11` will not work now that the mopa bootstrap bundle was removed.

PR #213, opened 2020-11-16
Upgrade to Symfony 4.4

mroest:


src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/config/security.yml
```diff
@@ -85,7 +85,7 @@ services:
 
     # Session
     self_service.security.authentication.session.session_storage:
-        public: false
+        public: true
```
MKodde, line 88: This certainly is the easiest way to achieve this, but consider using DI to get this service into the controller. That way this is more future proof.

app/AppKernel.php
```diff
@@ -28,7 +28,6 @@ public function registerBundles()
             new Symfony\Bundle\SecurityBundle\SecurityBundle(),
             new Symfony\Bundle\TwigBundle\TwigBundle(),
             new Symfony\Bundle\MonologBundle\MonologBundle(),
-            new Symfony\Bundle\AsseticBundle\AsseticBundle(),
```
MKodde, line 31: In addition to the removed code in this commit also please remove any other Assetic config references. I grepped and found these:

- https://github.com/OpenConext/Stepup-SelfService/blob/7a46375d0a2d83c3cac911114f07d9d3d0bd47fd/app/config/config_dev.yml#L36-L37
- https://github.com/OpenConext/Stepup-SelfService/blob/996ed6af12794b2492eeb7e3a6db327c810f3f57/composer.json#L30
- https://github.com/OpenConext/Stepup-SelfService/blob/5b7f50540b95756eabd4e5a9f6eda6855e5d2602/app/config/parameters.yml.dist#L11

app/AppKernel.php
```diff
@@ -43,7 +42,6 @@ public function registerBundles()
 
         if (in_array($this->getEnvironment(), array('dev', 'test'))) {
             $bundles[] = new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
-            $bundles[] = new Sensio\Bundle\DistributionBundle\SensioDistributionBundle();
```
MKodde, line 47: Did you leave it in the Composerfile on purpose?

https://github.com/OpenConext/Stepup-SelfService/blob/996ed6af12794b2492eeb7e3a6db327c810f3f57/composer.json#L25

.gitignore
```diff
@@ -1,9 +1,7 @@
-/web/bundles/
-/web/app_dev.php
+//bundles/
+/public/app_dev.php
```
MKodde, line 2: Shouldnt this be `/public/index.php` ?

MKodde (changes requested):
A real good starting point! See some feedback below.

src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/config/security.yml
mroest: Agree, but at this point since the changeset is already so big i do not want to change too much. Especially not refactorings that should go in a different PR. The setting was wrong before. Apparently SF did not do much with the public directive.

MKodde (approved):

PR #214, opened 2020-11-19
Allow HTML in explanation/description

mroest:
This will enable SURF to enrich the description used in emails with
links etc.

See: https://www.pivotaltracker.com/story/show/174901359


MKodde (approved):
I think you got them all

phavekes:
Tested, and works as expected
:shipit:

PR #220, opened 2021-02-23
Initial Remote Vetting cleanup

pablothedude:
Some cleanup needed to be done in order to remove some shortcuts made
during development of the RV POC.

This refactoring consists of:
- Remove double escaping from identity log
- Make the RV SP configuration static (acs endpoint and entityId)
- Remove double escaping from the RV process session data
- Use dedicated mock endpoint for testing instead of relying on Irma configuration
- Move the mock services out of the production service configuration
- Add assertion to the session test to get rid of risky tests
- Use a dedicated encryption key-pair in tests to ease maintenance
- Use the ApplicationHelper itself instead of a mock in the integration test to test it's behavior.


src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/config/remote_vetting.yml
```diff
@@ -5,7 +5,10 @@ services:
 
   Surfnet\StepupSelfService\SelfServiceBundle\Service\RemoteVetting\ServiceProviderFactory:
     arguments:
-      - '%remote_vetting_sp%'
+      $router: '@router'
+      $entityId: "%remote_vetting_entity_id%"
+      $assertionConsumerUrlSlug: "ss_second_factor_remote_vet_acs"
```
MKodde, line 11: Why hard-code it in the service container? Might be more predictable to hard code it in code?

config/legacy/parameters.yaml.dist
```diff
@@ -62,6 +62,8 @@ parameters:
 
     # remote vetting
 
+    remote_vetting_entity_id: https://selfservice.stepup.example.com/rv/metadata
+
```
MKodde, line 66: Maybe I lack context, but isn't the RV configuration not still configurable? Albeit it became less pronounced, but you can still configure the entitiy id in parameters.yml.

config/legacy/samlstepupproviders.yaml
```diff
@@ -83,32 +33,58 @@ surfnet_stepup_self_service_saml_stepup_provider:
                 pop_failed: "%gssp_webauthn_pop_failed%"
                 app_android_url: "%gssp_webauthn_app_android_url%"
                 app_ios_url: "%gssp_webauthn_app_ios_url%"
-        azuremfa:
```
MKodde, line 86: This might lead to issues when building the dev environment. The application provisioning first composer-installs with scripts using the VCS provided parameters. It then dumps the StepupDeploy specific parameters. And again composer installs them. IIRC the first run requires the azuremfa gssp to be in config. I do not have time to re-provision my SelfService at this moment. But maybe this warning rings a bell with you? It might well be, I'm mistaken here.

MKodde (commented):
I like what I see here! See some feedback in the code review section below. I think this is good to ship. But maybe you want to address some of my points. It's up to you.

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/RemoteVetting/Encryption/IdentityFilesystemWriter.php
```diff
@@ -40,7 +40,7 @@ public function __construct(RemoteVettingConfiguration $configuration)
     public function write($data)
     {
         $fileName = $this->createFileName($this->configuration->getLocation());
-        file_put_contents($fileName, json_encode($data));
+        file_put_contents($fileName, $data."\n");
```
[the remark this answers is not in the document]
pablothedude: The output files are now cross platform so I reckoned this would be much better.

src/Surfnet/StepupSelfService/SelfServiceBundle/Service/RemoteVetting/ServiceProviderFactory.php
```diff
@@ -28,14 +31,16 @@ class ServiceProviderFactory
      */
     private $serviceProvider;
 
-    /**
-     * @param array $configuration
-     */
-    public function __construct(array $configuration)
+    public function __construct(RouterInterface $router, string $entityId, string $assertionConsumerUrlSlug, string $privateKey)
```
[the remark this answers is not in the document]
pablothedude: I totally agree to add type hinting while at it. But I don't see the the reason why a constructor should be type hinted.

config/legacy/parameters.yaml.dist
pablothedude: Only the entityid and keys, but not the endpoint anymore, that seemed to be confusing because of the multiple acs endpoints that exist.

config/legacy/samlstepupproviders.yaml
pablothedude: I had to change the config in order to get my dedicated RV Stepup-VM up and running with the config in the remote-vetting branch in Deploy. Not all GSSP's were and will be configured for RV so this was done to just reflect that config. This indeed should be addressed to reflect develop in Deploy but I think this should be addressed just before (and only if) RV lands back in develop. But because that could happen in a while and a lot of Deploy's config could have been changed in the meantime I'll prefer to keep this config in sync with the RV branch in Deploy.

src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/config/remote_vetting.yml
pablothedude: The slug of the route itself isn't hardcoded either so the slug is dependency of the ServiceProviderFactory and that's why i used DI.

PR #221, opened 2021-02-24
Add GSSP UserAttributes extensions to a SAMLAuthnRequest

mroest:
Passing user attributes like an emailaddress enables GSSP providers
to ease registration of tokens. Ie. AzureMFA can use the emailaddr without
asking it for a 'second' time.

See: https://www.pivotaltracker.com/story/show/175121584


MKodde:
For a POC this seems to work as intended! Nice job adding the service to capture this business rule! This makes for a nice starting point to implement the actually requested rules specified in `https://www.pivotaltracker.com/story/show/175121584` in Step 2. Maybe this is good enough for an initial release, as this change set also limits exposing the user email only for Azure MFA.

MKodde (approved):
Approved, with the remark that this PR not yet fully covers the requirements specified in the story. The gssp typ to extension mapping is not yet configurable.

PR #227, opened 2021-03-17
Self vetting

MKodde:
This monster PR adds Self Vet support to the Self Service. This allows the users of institutions to self vet second factor tokens. Given they have at least one RA-vetted token.

I tried to make the commits somewhat atomic, but failed in doing so on several occasions as certain parts of the affected files where refactored a couple of times over to get good alignment with the Remote Vetting solution. And to improve certain aspects of the code in general. Still I suggest you review this on a per-commit basis.

https://www.pivotaltracker.com/story/show/176715448


src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SelfVetController.php
```diff
@@ -0,0 +1,209 @@
...
+                    )
+                );
+                throw new AuthenticationException('Unexpected InResponseTo in SAMLResponse');
+            }
+            $candidateSecondFactor = $this->secondFactorService->findOneVerified($secondFactorId);
+            // Proof of possession of higher/equal LoA was successful, now apply the self vet command on Middleware
+            $command = new SelfVetCommand();
+            $command->identity = $this->getIdentity();
+            $command->secondFactor = $candidateSecondFactor;
+            $command->authoringLoa = $assertion->getAuthnContextClassRef();
```
pablothedude, line 197: Why not send the authorizing secondFactorID, so the LOA could be validated in MW because there is were the domain logic should be validated. Consider also sending the authoring LOA so MW could validate if the vetted token has the LOA as expected by MW.

src/Surfnet/StepupSelfService/SelfServiceBundle/Tests/Service/GsspUserAttributeServiceTest.php
```diff
@@ -48,12 +48,6 @@ public function setUp(): void
         parent::setUp();
     }
 
-    public function tearDown(): void
```
pablothedude, line 51: :+1: I removed the teardown also from RV but hadn't it backported yet. So thanks!

pablothedude (changes requested):
Nice work, and a valuable addition to the platform!

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SelfVetController.php
MKodde: I do not know how to retrieve the token used in the gateway authentication. There is a possibility the users possesses multiple suitable SF tokens, and the WAYG is used to choose to token of his/her liking. The assertion only displays the LOa level that was reached, not the Second Factor ID. Unless maybe this can be fetched from a magic session I'm not aware of but you might?

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
```diff
@@ -77,10 +78,18 @@ public function consumeAssertionAction(Request $httpRequest)
...
 
-        $logger->notice('Received an authentication response for testing a second factor');
-
         $session = $this->get('session');
-
+        // The test authentication IdP is also used for self vetting, a different session id is
+        // used to mark a self vet command
+        if ($session->has(SelfVetController::SELF_VET_SESSION_ID)) {
+            /** @var SelfVetRequestId $selfVetRequestId */
+            $selfVetRequestId = $session->get('second_factor_self_vet_request_id');
```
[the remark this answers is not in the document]
MKodde: Good catch, updating it right away.

public/index.php
```diff
@@ -19,6 +19,10 @@
 if ($trustedHosts = $_SERVER['TRUSTED_HOSTS'] ?? false) {
     Request::setTrustedHosts([$trustedHosts]);
 }
+if (isset($_COOKIE['testcookie']) && strpos($_SERVER['HTTP_USER_AGENT'], 'GuzzleHttp') !== false) {
```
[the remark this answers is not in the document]
MKodde: I made it into this:

```
// Support for behat test from stepup-deploy, when the `testcookie` and GuzzleHttp user agent
// are present and app_env is not prod, then engage smoketest mode.
if ($_SERVER['APP_ENV'] !== 'prod' &&
    isset($_COOKIE['testcookie']) &&
    strpos($_SERVER['HTTP_USER_AGENT'], 'GuzzleHttp') !== false
) {
    $_SERVER['APP_ENV'] = 'smoketest';
    $_SERVER['APP_DEBUG'] = true;
}
```

Not certain this is the best solution. Maybe a dedicated dev-index.php provisioned only for dev machines might be a more robust solution. For now this should at least keep prod envs out of the smoketest mode.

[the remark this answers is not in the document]
pablothedude: I think this would suffice for now. Personally I'm more fan of a single index.php. So you always see the same file as production when working on it.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SelfVetController.php
pablothedude: I was under the assumption that you would have to select the token you would authorize the token with in SA. That way you would have a decent audit trail and MW would be the SSOT. Do you know the motives why this wasn't done?

public/index.php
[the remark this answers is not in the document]
MKodde: I think we are going to implement that in the near future as part of the `fragments` - `secret` sec updates.

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SelfVetController.php
MKodde: That might have been discussed at some point, but re-using the existing WAYG in Gateway was decided on. 

1. Faster/easier to develop
2. Prevent code duplication

pablothedude: Clear, but be aware that this will be hard to change later on due to the event sourced MW.

pablothedude (approved):

src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/views/registration/display_vetting_types.html.twig
```diff
@@ -52,5 +38,19 @@
               </div>
           </div>
         {% endif %}
+        <div class="col-xs-12 col-md-3">
+            <div class="thumbnail remote-vetting-selector">
```
MKodde, line 42: I think changing the css selector to vetting-selector would be more fitting. That way all other vetting types would also apply.

public/scss/application.scss
```diff
@@ -25,6 +25,24 @@
     > div:nth-child(3n+4) {
       clear: left;
     }
+    .remote-vetting-selector{
```
MKodde, line 28: Considering my previous review entry. Removing the 'remote-' portion from this scss decoration should be all it takes.

MKodde (commented):
Hi @phavekes thanks for your contribution. The changes looked very good. Had some small nitpicks you might want to address

PR #240, opened 2021-10-04
Remove initiate screens for GSSP's

MKodde:
The initiate screen is not required as it is now only an extra click in the registration flow. Note that this is only done for the GSSP tokens. The registration action requires a POST request, so this was implemented on the generic_second_factor.html.twig.

The route is renamed to be more status focussed. It is now only used to come back to when registration failed on the GSSP side.

https://www.pivotaltracker.com/story/show/179482160


src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/views/registration/partial/generic_second_factor.html.twig
```diff
@@ -25,9 +25,11 @@
...
         </p>
         <p>
-            <a href="{{ url }}" class="btn btn-primary">
-                {{ secondFactor.getButtonUse() }}
-            </a>
+            <form method="post" enctype="text/plain" accept-charset="UTF-8" action="{{ url }}">
+                <button type="submit" class="btn btn-primary">
+                    {{ secondFactor.getButtonUse() }}
+                </button>
+            </form>
```
MKodde, line 32: The route that jumps to the GSSP only allows POST actions that is why we moved away from a simple anchor.  UI wise this looked identical to the anchor solution. This should function the same way from a UX perspective too.

src/Surfnet/StepupSelfService/SelfServiceBundle/Value/GsspToken.php
```diff
@@ -63,7 +63,7 @@ private function __construct(ViewConfig $viewConfig, $type)
      */
     public function getRoute()
     {
-        return 'ss_registration_gssf_initiate';
+        return 'ss_registration_gssf_authenticate';
```
MKodde, line 66: The heart of this change lies here. The route is updated to directly reference the `ss_registration_gssf_authenticate` action. Skipping the initiate route.

src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/views/registration/partial/generic_second_factor.html.twig
Badlapje: ❤️ the explanation 👍

Badlapje (approved):
Looks like great work to me.  Much obliged for the clarificatory comments!

PR #246, opened 2022-01-31
Create tag-release.yml

MKodde:
This workflow will start a release procedure using Stepup-Build scripts. Resulting in a tarbal and shasum that is uploaded to the github release page


phavekes (approved):

MKodde:
I updated the changelog, tagged 3.5.4 and ran a final test of the build scritpt with that tag. All seemed well! Merging

PR #247, opened 2022-02-08
Enhance logging

tvdijen:


MKodde (approved):
Thanks Tim, making things more specific sounds like a plan!

Having the full response body in the logs is no problem from my point. I rarely have to audit this territory. But having the response body seems valuable debug information.

PR #261, opened 2022-06-21
Integrate Recovery tokens in SelfService

MKodde:
This PR creates the self-asserted token registration feature into SelfService. It continues the work done on #260 

An effort was put into making the commits of this PR somewhat atomic, but that did not work out very well. 

The tasks found in the base story will give a more clear understanding of what work is done on this PR.

https://www.pivotaltracker.com/story/show/181934012

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SelfAssertedTokensController.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SelfAssertedTokensController.php
@@ -209,6 +209,8 @@ class SelfAssertedTokensController extends Controller
 
         $command = new VerifySmsRecoveryTokenChallengeCommand();
         $command->identity = $identity->id;
+        $command->resendRoute = 'ss_second_factor_self_asserted_tokens_recovery_token';
+        $command->resendRouteParameters = ['secondFactorId' => $secondFactorId, 'recoveryTokenId' => $recoveryTokenId];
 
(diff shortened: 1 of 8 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Form/Type/AuthenticateSafeStoreType.php
```diff
@@ -0,0 +1,54 @@
...
+use Symfony\Component\Form\Extension\Core\Type\PasswordType;
+use Symfony\Component\Form\Extension\Core\Type\SubmitType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class AuthenticateSafeStoreType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder->add('secret', PasswordType::class, [
```
pablothedude, line 32: Does the password field needs a solution so the input text or last character can become visible? ASVS 2.1.12

pablothedude (approved):

src/Surfnet/StepupSelfService/SelfServiceBundle/Form/Type/AuthenticateSafeStoreType.php
MKodde: Thanks, great suggestion! I will discuss this with the PM's.

PR #266, opened 2022-07-19
Allow expression of a preferred activation flow (vetting type)

MKodde:
SelfService now listens for a query string on the entry point action. (`/`). The name of this QS field and the available options can be configured in parameters.yaml.

The value of the QS param value can manipulate the way the vetting types are rendered during token activation. 

See: https://www.pivotaltracker.com/story/show/181966790


MKodde:
image

pablothedude (approved):

Issue #271, opened 2022-10-25
4.0: Feature request: completely disable SMS recovery tokens


tvdijen:
I'm not able to work with either Spryng or Messagebird, so I would like to be able to disable SMS recovery tokens completely. A feature toggle would be very helpful!

MKodde:
It makes sense to allow disabling one of the recovery token types. Not using SMS for one is a very legitimate reason for you require it to be disabled. This never crossed our minds while building this feature.

I opened #274 addressing this issue.

tvdijen:
Was closed in #274

PR #274, opened 2022-11-01
Add toggles to disable recovery token types

MKodde:
SMS or safe store recovery token types can be enabled and disabled. But never can both be disabled.

See https://www.pivotaltracker.com/story/show/183636781

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -7629,16 +7629,16 @@
         },
         {
             "name": "twig/twig",
-            "version": "v2.15.1",
+            "version": "v2.15.3",
             "source": {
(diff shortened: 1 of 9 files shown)


MKodde:
Would this work for you @tvdijen in the light of https://github.com/OpenConext/Stepup-SelfService/issues/271?

tvdijen:
I wonder if it makes more sense to add this to the `middleware-institution.json` as `allowed_recovery_methods`.. This touches corporate policies and may differ per org.

Shaky212 (approved):

PR #276, opened 2022-12-08
Improve verification mail message timing

MKodde:
To reach this goal, we use the new command that sends the registration
mail. And send it when the ra-vetting vetting type was selected by the
Identity.

An intermediate action was added to the registration flow that only
sends the mail message and then forwards to the existing 'registration
mail message sent' action. This prevents the re-sending of the mail
message on page reloads of the 'sent' page.

See: https://www.pivotaltracker.com/story/show/183826628


Shaky212 (approved):
Looks good!

PR #281, opened 2023-01-23
Skip the SMS authentication step during SAT registration

MKodde:
The authentication step is considered redundant when you already proved possession by filling an SMS OTP.


Shaky212 (approved):
Looks good!

PR #295, opened 2023-06-06
Make sms form autocomplete="one-time-code"

thijskh:
This allows the (mobile) browser to help autofill


phavekes (approved):

MKodde (approved):

MKodde:
After merging this, I discussed this over some java with one of our front-end experts. He confirmed that this feature should be safe to use. Browsers should fall back on default behavior when they do not understand `autocomplete="one-time-code"`. What their default value would be is not clear to me for every browser. But the major (mobile) browsers will support this just fine.

PR #299, opened 2023-10-05
Remove unnecessary enctype from form

tvdijen:
OWASP core ruleset will complain about this, because the use of `text/plain` in conjunction with `method=post` is effectively useless (form values will never become available in the $_POST global).

The form functions just fine without it.


MKodde (approved):
Thanks for this improvement @tvdijen 

Merging. This should land in the next release

PR #303, opened 2023-11-06
Update Docker configs

quartje:
Add docker configuration


.github/workflows/build-push-docker-image.yml
```diff
@@ -0,0 +1,58 @@
+name: build-push-docker-image
+
+#on: workflow_dispatch
```
MKodde, line 3: Must this be re-enabled?

docker/conf/apache2.conf
```diff
@@ -0,0 +1,35 @@
+<Virtualhost *:80>
+    ServerName  selfservice
+    ServerAdmin admin@dev.openconext.local
+
+    DocumentRoot /var/www/html/public
+
+    SetEnv HTTPS on
+    SetEnv APP_ENV prod
```
MKodde, line 8: Can this be removed? Setting it in the Apache config makes overriding this var quite hard.

quartje: Yes, you're absolutely right. I've removed it now!

MKodde (approved):

PR #305, opened 2023-11-14
Replace forms with a-href links

tvdijen:
When setting a Content-Security-Policy with `form-action *.<your-domain>.nl`, the forms won't work in Edge/Chrome.
See [this](https://content-security-policy.com/form-action/) article, paragraph "Can the form-action redirect to another url?" for the reason behind this. TL;DR: Edge/Chrome does not allow the form's POST-destination to perform a redirect.
Replacing the unnecessary forms with simple links resolves this issue and allows for a stricter CSP-policy.

Previously discussed on Slack with @MKodde

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/views/registration/display_vetting_types.html.twig
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/views/registration/display_vetting_types.html.twig
@@ -39,9 +39,9 @@
                   <h3>{{ 'ss.registration.vetting_type.title.self_vet'|trans }}</h3>
                   <hr>
                   <p>{{ 'ss.registration.vetting_type.description.self_vet'|trans }}</p>
-                  <form action="{{ url('ss_second_factor_self_vet', {'secondFactorId': secondFactorId}) }}" method="get">
+                  <a href="{{ url('ss_second_factor_self_vet', {'secondFactorId': secondFactorId}) }}">
                       <button type="submit" class="btn btn-primary">{{ 'ss.registration.vetting_type.button.self_vet'|trans }}</button>
-                  </form>
+                  </a>
               </div>
           </div>
         {% endif %}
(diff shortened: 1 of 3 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/views/registration/display_vetting_types.html.twig
```diff
@@ -66,9 +66,9 @@
                 <h3>{{ 'ss.registration.vetting_type.title.ra_vetting'|trans }}</h3>
                 <hr>
                 <p>{{ 'ss.registration.vetting_type.description.ra_vetting'|trans }}</p>
-                <form action="{{ url('ss_registration_send_registration_email', {'secondFactorId': secondFactorId}) }}" method="get">
+                <a href="{{ url('ss_registration_send_registration_email', {'secondFactorId': secondFactorId}) }}">
                     <button type="submit" class="btn btn-primary">{{ 'ss.registration.vetting_type.button.ra_vetting'|trans }}</button>
```
MKodde, line 70: This construction is not 100% yet from a html structure point of view.

I'd say: remove the `<button>` element but keep the button text as the anchor text for the `<a>` element. And style the `<a>` to make it look like a button. I think we already have utility (css) classes ready for that.

Would you be so kind to look into that?

templates/base.html.twig
```diff
@@ -45,12 +45,12 @@
         </div>
         {% if app.user %}
             <div class="clearfix page-header-user">
-                <form method="post" action="{{ logout_url('saml_based') }}" class="pull-right">
+                <a href="{{ logout_url('saml_based') }}" class="pull-right">
```
MKodde, line 48: See previous feedback point.

MKodde (commented):
Conceptually this makes lots of sense. Not sure about the logout button tho, or did that also trigger the same CSP error?

Given the fact we are not actually logging out, this is not a big issue in my opinion.

tvdijen:
Logout got me the same CSP-error yes. Our logout-url is on another domain.

src/Surfnet/StepupSelfService/SelfServiceBundle/Resources/views/registration/display_vetting_types.html.twig
tvdijen: Sorry, I spent the entire afternoon but I can't get it to look visually the same without the button-tags. Leaving it 'as-is', but feel free to try it yourself

templates/base.html.twig
thijskh, line 48: I don't think this will work because it changes the POST to a GET, and selfservice does not allow GET requests to the logout endpoint.

tvdijen: That's cheating, because the GET-method was only removed two weeks ago ;)

thijskh: Not quite, in 4.0.9 (released Jun 6, 2023) I can also not use GET for logout, it only accepts POST.

thijskh (changes requested):
As noted in the comment, this appears to break logout when merged as-is

PR #309, opened 2023-12-07
Build and publish test containers

quartje:
This PR adds a Github action for building a test container, which will be used in behat tests


MKodde (approved):
I reviewed @quartjes changes predominantly and also looked at my own changes. I suggest another reviewer looks at this too.

parijke (approved):
Nul opmerkingen, alleen vraag:

Waarom is nu bij forms name nodig en voorheen niet? Is dat esthetisch of heeft het ook een functie?

MKodde:
> Why are forms now set with a `name` attribute.

This is purely for testing purposes. Maybe using a data-test attribute would have been more in line with the solution I chose later down the line. But still. I'm happy with this as-is.

PR #313, opened 2024-03-04
Ensure display of Authn errors

MKodde:
In SF 6, the new exception handler does not show the exception message in the exception handler. Rather, these errors are handled at another point (authn listeners).

Also see: https://stackoverflow.com/questions/40174266/not-all-exceptions-catched-within-the-error-handler


parijke (approved):

parijke:
@thijskh approved

PR #315, opened 2024-03-11
Simplify exception logical condition, showing the attribute missing error message once again

MKodde:
When rendering the custom exception page, we test if the MissingRequiredAttributeException. But it also had to be a HTTP exception. That was never the case (also before #313).

By removing the first part of the statement, the error message is displayed correctly once again.

See: https://github.com/OpenConext/Stepup-SelfService/pull/315


MKodde:
Now, when an attr is missing we are treated with this error message:

image

parijke (approved):

PR #320, opened 2024-04-16
Set correct SelfVet ACS route

MKodde:
The additional SelfVet namespace was not yet reflrected in the route (forward) name.

This was the only occurence where the SelfVetController is called by name. There should not be other wrong calls in the project.

See: https://www.pivotaltracker.com/story/show/187412638

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,16 @@
 # Changelog
 
+# 5.0.4
+- Correct a route name with a missing namespace part (self vetting acs) #320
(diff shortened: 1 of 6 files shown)


parijke (approved):
I see I introduced this bug while refactoring the controllers? Anyway, good catch @MKodde

MKodde:
There references are quite invisible unless you run the code. And we can conclude that a GSSP self-vet flow is not part of our Behat test suite.

PR #323, opened 2024-05-22
Remove framework.session.*liftime settings

MKodde:
They make the app cookie into a session cookie with a given lifetime. We want them to be a non persistent cookie

See: https://www.pivotaltracker.com/story/show/187651480


MKodde:
Closing this initial work. The real solution can be found in #325. Thanks @parijke for fixing this!

PR #325, opened 2024-05-23
Programatically ensure cookie expiration

parijke:
This add listeners to the chain to detect and set cookie settings. Cherrypicked #323 so that one can be deleted.

### DIFF
--- a/config/packages/framework.yaml
+++ b/config/packages/framework.yaml
@@ -19,8 +19,9 @@ framework:
         cookie_httponly: true
         cookie_secure: true
         cookie_samesite: none
-        cookie_lifetime: '%session_max_absolute_lifetime%'
-        gc_maxlifetime: '%session_max_relative_lifetime%'
+        cookie_lifetime: 0
+        gc_maxlifetime: 0
+
 
     fragments: false
     error_controller: Surfnet\StepupSelfService\SelfServiceBundle\Controller\ExceptionController::show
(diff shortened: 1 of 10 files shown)


config/packages/framework.yaml
```diff
@@ -19,8 +19,9 @@ framework:
         cookie_httponly: true
         cookie_secure: true
         cookie_samesite: none
-        cookie_lifetime: '%session_max_absolute_lifetime%'
-        gc_maxlifetime: '%session_max_relative_lifetime%'
+        cookie_lifetime: 0
+        gc_maxlifetime: 0
```
MKodde, line 23: Do you  intentionally configure the cookie_lifetime and gc_maxlifetime? Do we need them configured this way?

In the original situation we relied on the default values for these settings (as in, we did not configure them)

Please double check this, as my memory does not always serve me right..

src/Surfnet/StepupSelfService/SelfServiceBundle/EventListener/ExplicitSessionTimeoutListener.php
```diff
@@ -0,0 +1,102 @@
...
+        private LoggerInterface                  $logger,
+        private EventDispatcherInterface         $eventDispatcher,
+    ) {
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            // The firewall, which makes the token available, listens at P8
+            // We must jump in after the firewall, forcing us to overwrite the translator locale.
```
MKodde, line 53: Seems like a comment `ddp`'ed from another class. At least the mention of the translator seems unrelated :man_shrugging:

src/Surfnet/StepupSelfService/SelfServiceBundle/Controller/SamlController.php
MKodde, line 1: Why did you cherry-pick this merely cosmetic commit? #999c46035f5

MKodde (approved):
Nice work on fixing the cookie lifetime issue :parrot:! Feel free to move ahead with this, but please also consider my feedback below. And our IRL discussion about the commit history.

PR #346, opened 2024-11-25
Use entitlement when no query parameter is set

pablothedude:
This change will use the only activation flow that's allowed through the entitlement SAML attribute when no preferred activation flow is set with a query parameter.

This was encountered while enhancing the Behat tests with the ability to add SAML-attributes.

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Service/ActivationFlowService.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Service/ActivationFlowService.php
@@ -83,6 +83,11 @@ class ActivationFlowService
             ];
         }
 
+        if ($requestedActivationPreference instanceof ActivationFlowPreferenceNotExpressed && count($availableActivationPreferences) === 1) {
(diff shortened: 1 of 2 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Service/ActivationFlowService.php
```diff
@@ -83,6 +83,11 @@ public function getPreference(): ActivationFlowPreferenceInterface
             ];
         }
 
+        if ($requestedActivationPreference instanceof ActivationFlowPreferenceNotExpressed && count($availableActivationPreferences) == 1) {
```
MKodde, line 86: Does PHPStan not complain about the unnecessarily vague expression here? 
```suggestion
        if ($requestedActivationPreference instanceof ActivationFlowPreferenceNotExpressed && count($availableActivationPreferences) === 1) {
```

MKodde (approved):
Approved with a suggestion to make the logical expression more expressive

Issue #440, opened 2024-12-02
Email validation link lost after login


pmeulen:
The email validation in SelfService works by sending a mail to the user's email address with a link in the form: https://sa.test.surfconext.nl/verify-email?n=337c06301b71d9ed4e2452969d737961

Steps to reproduce:
0. Login to selfservice with an account from an institution with email validation enabled (e.g. joe-a1)
1. Add a new token
2. Open a new private window or otherwise remove the selfservice session cookie and open the verify-email link in that window
3. You need to authenticate, if asked use the same account (joe-a1)

Expected behaviour: The email address is validated and the user arrives at e.g. the https://sa.test.surfconext.nl/second-factor/ page. Selfservice remembers the link that was used before started the SAML authentication and redirects the user back to that link. This must happen for all links that trigger a new authentication: both when no session exists and when the session is expired. (#360)
What happens: After authentication the user is directed to the home page (https://sa.test.surfconext.nl/overview). The link is lost. This is confusing users who click the link in the email and are taken to the selfservice homepage as nothing happens.
 
 selfservice version: 5.0.4p1
 
 Please add tests for this scenario.

pablothedude:
Seems related to: https://github.com/OpenConext/Stepup-SelfService/pull/346

PR #443, opened 2024-12-05
Redirect to initial uri after successful login

pablothedude:
https://github.com/OpenConext/Stepup-SelfService/issues/440

Also added a test in devconf: https://github.com/OpenConext/OpenConext-devconf/pull/22

### DIFF
--- a/src/Surfnet/StepupSelfService/SelfServiceBundle/Security/Authentication/Handler/AuthenticationSuccessHandler.php
+++ b/src/Surfnet/StepupSelfService/SelfServiceBundle/Security/Authentication/Handler/AuthenticationSuccessHandler.php
@@ -37,8 +37,8 @@ class AuthenticationSuccessHandler extends SuccessHandler
 
     public function onAuthenticationSuccess(Request $request, TokenInterface $token): ?Response
     {
-        $this->authenticatedSessionStateHandler->setCurrentRequestUri($request->getUri());
+        $redirectUri = $this->authenticatedSessionStateHandler->getCurrentRequestUri();
 
-        return new RedirectResponse('/');
+        return new RedirectResponse($redirectUri);
     }
 }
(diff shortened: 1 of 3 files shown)


src/Surfnet/StepupSelfService/SelfServiceBundle/Security/Authentication/Handler/AuthenticationSuccessHandler.php
```diff
@@ -37,8 +37,8 @@ public function __construct(
 
     public function onAuthenticationSuccess(Request $request, TokenInterface $token): ?Response
     {
-        $this->authenticatedSessionStateHandler->setCurrentRequestUri($request->getUri());
+        $redirectUri = $this->authenticatedSessionStateHandler->getCurrentRequestUri();
```
johanib, line 40: Is no fallback needed?
It is a given that the URL is always set?

In theory it might return an empty string. Maybe fallback to '/' in that case?

johanib (commented):
Is there no test coverage?

---

Did you come up with this? Or did you reverse engineer from before the symfony upgrade?

src/Surfnet/StepupSelfService/SelfServiceBundle/Security/Authentication/Handler/AuthenticationSuccessHandler.php
pablothedude: This should be always set in: `\Surfnet\StepupSelfService\SelfServiceBundle\Security\Authentication\SamlAuthenticator::start`
Or: \Surfnet\StepupSelfService\SelfServiceBundle\EventListener\ExplicitSessionTimeoutListener::checkSessionTimeout

pablothedude: I howevere the fallback in the SessionStorage so this should be handled if this unexpectedly the case.

johanib (approved):

